PR #4, opened 2015-02-20
Update parent POM to the latest.

jglick:
@reviewbybees in preparation for a release of 1.2.


stephenc:
:+1: though where is the pull request builder and are there any fun jelly escape issues to bite you

jglick:
> where is the pull request builder

Do not think we have one for this project yet. Anyway the tests only take a few seconds to run.

> are there any fun jelly escape issues

That is a different parent POM you are thinking of, for Jenkins plugins. This is just a generic library POM.

recampbell:
:+1:

PR #2, opened 2015-02-18
Memory leak

jglick:
Found that `Caller.Info` can be holding onto script classes even after the script has finished. @reviewbybees esp. @kohsuke

PR #6, opened 2015-02-20
Null check for synchronous calls

jglick:
Fixes regression in #2:

```
java.lang.NullPointerException
    at com.cloudbees.groovy.cps.impl.Caller.isAsynchronous(Caller.java:50)
    at com.cloudbees.groovy.cps.impl.Caller$isAsynchronous.call(Unknown Source)
    at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.each(CpsDefaultGroovyMethods.groovy:24)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at org.codehaus.groovy.runtime.metaclass.ReflectionMetaMethod.invoke(ReflectionMetaMethod.java:51)
    at org.codehaus.groovy.runtime.metaclass.NewInstanceMetaMethod.invoke(NewInstanceMetaMethod.java:54)
    at org.codehaus.groovy.runtime.callsite.PojoMetaMethodSite$PojoMetaMethodSiteNoUnwrapNoCoerce.invoke(PojoMetaMethodSite.java:271)
    at org.codehaus.groovy.runtime.callsite.PojoMetaMethodSite.call(PojoMetaMethodSite.java:53)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:116)
    at jenkins.model.InterruptedBuildAction.summary$_run_closure1.doCall(summary.groovy:6)
```

This seems to have been because `record` was not yet called in this thread, so the initial `Info` had null fields, which used to be OK but now I am calling `i.receiver.get()` rather than merely `i.receiver`.

I struggled hard to make `CallerTest` reproduce the realistic call stack from `CpsDefaultGroovyMethods`, by creating a plain `GroovyShell` and making it run the same script from `CpsTransformerTest.each`, but I could not get it to fail—though the original `each` test would fail, without having modified its source code! (It would claim the sum was 0 rather than 55, as if the `each` loop had not run at all.) I think this is because all the tests run in the same JVM. I tried using a different JVM per test suite, but `CallerTest` still passed. I made sure that the static initializer in `ContinuationGroup` was being called, so that `CpsDefaultGroovyMethods` would be registered in the VM-wide metamodel for `each`, but Groovy insisted on calling the standard version in `DefaultGroovyMethods` directly. (`CpsDefaultGroovyMethods` falls back to this if `isAsynchronous` returns false.) Creating a new test extending `AbstractGroovyCpsTest`, or even copying the minimal parts of that over, did not work at all, I think because various parts of `CpsTransformer` call `each` on various lists. So I gave up and just tested `Caller.isAsynchronous` directly.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/groovy/cps/impl/Caller.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/Caller.java
@@ -29,7 +29,7 @@ public class Caller {
      */
     public static boolean isAsynchronous(Object receiver, String method, Object... args) {
         Info i = store.get();
-        return receiver==i.receiver.get() && method.equals(i.method) && arrayShallowEquals(i.args, args);
+        return i.receiver != null && receiver==i.receiver.get() && method.equals(i.method) && arrayShallowEquals(i.args, args);
     }
 
     private static boolean arrayShallowEquals(Reference<Object>[] a, Object[] b) {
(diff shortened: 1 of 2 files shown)


tfennelly:
:+1: 

insofar as I get it

PR #10, opened 2015-03-13
[JENKINS-27271] Implement safe navigation operator

jglick:
[JENKINS-27271](https://issues.jenkins-ci.org/browse/JENKINS-27271)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java
@@ -36,10 +36,13 @@ public class FunctionCallBlock implements Block {
 
     private final SourceLocation loc;
 
-    public FunctionCallBlock(SourceLocation loc, Block lhsExp, Block nameExp, Block[] argExps) {
+    private final boolean safe;
+
+    public FunctionCallBlock(SourceLocation loc, Block lhsExp, Block nameExp, boolean safe, Block[] argExps) {
         this.loc = loc;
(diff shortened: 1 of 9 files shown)


jtnord:
:+1:

src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallBlock.java
```diff
@@ -97,7 +100,11 @@ private Next dispatchOrArg() {
                     return k.receive(v);
                 } else {
                     // regular method call
-                    return methodCall(e,loc,k,lhs,name,args);
+                    if (safe && lhs == null) {
+                        return k.receive(null);
```
kohsuke: Check this late implies that `x?.foo(a())` would evaluate `a()`. Is that what Groovy actually does? I think it might skip evaluating arguments altogether if `lhs` evaluates to `null`.

jglick:
```
$ groovy -e 'def x = null; def y = null; println x?.foo(y.bar)'
Caught: java.lang.NullPointerException: Cannot get property 'bar' on null object
java.lang.NullPointerException: Cannot get property 'bar' on null object
	at script_from_command_line.run(script_from_command_line:1)
```

so I think that is what Groovy actually does, for better or for worse.

kohsuke:
We could have had a test for that case, but let's just merge it.

:+1:

PR #13, opened 2015-06-20
Added a hook to insert a subtype of CpsClosure.

kohsuke:
Allow the calling code to use a subtype of `CpsClosure` to customize the behaviour of closure.

This was necessary for Jenkins Workflow plugin to intercept the sleep()
call, which requires adding a declared method on CpsClosure.

See more about this in jenkinsci/workflow-plugin #131.

In anticipation of additional CPS transformer configuration switches in
the future, this change introduces a class that collects such random
switches.

### DIFF
--- a/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
+++ b/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
@@ -16,6 +16,7 @@ import org.codehaus.groovy.control.customizers.CompilationCustomizer
 import org.codehaus.groovy.runtime.powerassert.SourceText
 import org.codehaus.groovy.syntax.Token
 
+import javax.annotation.Nonnull
 import java.lang.annotation.Annotation
 import java.lang.reflect.Modifier
 
@@ -82,7 +83,7 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor 
         super(CompilePhase.CANONICALIZATION)
     }
 
-    public void setConfiguration(TransformerConfiguration config) {
+    public void setConfiguration(@Nonnull TransformerConfiguration config) {
         this.config = config;
     }
 
--- a/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
+++ b/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
@@ -76,9 +76,14 @@ import static org.codehaus.groovy.syntax.Types.*
 class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor {
     private int iota=0;
     private SourceUnit sourceUnit;
+    private TransformerConfiguration config = new TransformerConfiguration();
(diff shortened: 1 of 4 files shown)


kohsuke:
@reviewbybees

stephenc:
:bee: not fully understanding all of this but I cannot see any bugs or increase in technical debt in this. It would be good to have some more javadoc comments though

src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
```diff
@@ -76,11 +76,16 @@ import static org.codehaus.groovy.syntax.Types.*
 class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor {
     private int iota=0;
     private SourceUnit sourceUnit;
+    private TransformerConfiguration config = new TransformerConfiguration();
 
     CpsTransformer() {
         super(CompilePhase.CANONICALIZATION)
     }
 
+    public void setConfiguration(TransformerConfiguration config) {
```
tfennelly: Seems like this should be `@NonNull`?

src/main/java/com/cloudbees/groovy/cps/TransformerConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+ * @see CpsTransformer#setConfiguration(TransformerConfiguration)
+ */
+public class TransformerConfiguration {
+    private ClassNode closureType = new ClassNode(CpsClosure.class);
+
+    public ClassNode getClosureType() {
+        return closureType;
+    }
+
+    public TransformerConfiguration withClosureType(ClassNode closureType) {
```
tfennelly: Maybe pass a `Class<? extends CpsClosure>` instead and create the `ClassNode` from it?

tfennelly: Ah sorry... I see you have an overloaded version.

src/main/java/com/cloudbees/groovy/cps/TransformerConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+        this.closureType = closureType;
+        return this;
+    }
+
+    public TransformerConfiguration withClosureType(Class c) {
```
tfennelly: Why not `Class<? extends CpsClosure>`?

src/main/java/com/cloudbees/groovy/cps/TransformerConfiguration.java
tfennelly: Why have this version of the method at all?

kohsuke: I think it's useful to define this fully general version as the base contract, because having that version supports all the use cases this library can support without frustrating users that "gee given the way this library is implemented I know it can take `ClassNode` but it just doesn't expose it!"

PR #19, opened 2015-11-25
Added SourceLocation.toString override

jglick:
For debugging https://github.com/jenkinsci/workflow-plugin/pull/249.

@reviewbybees esp. @kohsuke


oleg-nenashev:
:bee:

PR #20, opened 2015-12-14
[JENKINS-32062] Arrays could not be passed directly to methods

jglick:
[JENKINS-32062](https://issues.jenkins-ci.org/browse/JENKINS-32062)

@reviewbybees esp. @kohsuke

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,12 @@
       <version>4.11</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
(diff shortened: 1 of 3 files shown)


kohsuke:
I see, so it was unexpected varargs expansion that was getting in the way.

:bee:

jglick:
Yes, the other calls to this constructor come from Java code, which _statically_ resolves a varargs or non-varargs call based on declared types. But the generated code making this call is done by Groovy, which seems to try to guess at the call mode based on the runtime type of its argument, bless its stochastic little heart.

oleg-nenashev:
:bee:

PR #21, opened 2016-03-28
[JENKINS-28277] NPE in CurriedClosure.<init>

kohsuke:
[JENKINS-28277](https://issues.jenkins-ci.org/browse/JENKINS-28277)

This issue got extra visibility in https://news.ycombinator.com/item?id=11362058

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/groovy/cps/Builder.java
+++ b/src/main/java/com/cloudbees/groovy/cps/Builder.java
@@ -153,8 +153,8 @@ public class Builder {
         return b;
     }
 
-    public Block closure(int line, List<String> parameters, Block body) {
-        return new ClosureBlock(loc(line),parameters,body,closureType);
+    public Block closure(int line, List<Class> parameterTypes, List<String> parameters, Block body) {
+        return new ClosureBlock(loc(line),parameterTypes,parameters,body,closureType);
     }
 
     public LValueBlock localVariable(String name) {
--- a/src/main/java/com/cloudbees/groovy/cps/impl/ClosureBlock.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/ClosureBlock.java
@@ -15,13 +15,15 @@ import java.util.List;
  */
 public class ClosureBlock implements Block {
     private final List<String> parameters;
+    private final List<Class> parameterTypes;
     private final Block body;
     // this field would be null if we are deserializing from data saved by old version
     private final Class<? extends CpsClosure> closureType;
     private final SourceLocation loc;
 
-    public ClosureBlock(SourceLocation loc, List<String> parameters, Block body, Class<? extends CpsClosure> closureType) {
+    public ClosureBlock(SourceLocation loc, List<Class> parameterTypes, List<String> parameters, Block body, Class<? extends CpsClosure> closureType) {
         this.loc = loc;
+        this.parameterTypes = parameterTypes;
         this.parameters = parameters;
         this.body = body;
         this.closureType = closureType;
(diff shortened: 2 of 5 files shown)


src/main/java/com/cloudbees/groovy/cps/Builder.java
```diff
@@ -153,8 +153,8 @@ public Block sequence(Block b) {
         return b;
     }
 
-    public Block closure(int line, List<String> parameters, Block body) {
-        return new ClosureBlock(loc(line),parameters,body,closureType);
+    public Block closure(int line, List<Class> parameterTypes, List<String> parameters, Block body) {
```
oleg-nenashev: Do we require a binary compatibility here? I suspect so

src/main/java/com/cloudbees/groovy/cps/impl/ClosureBlock.java
```diff
@@ -15,13 +15,15 @@
...
 public class ClosureBlock implements Block {
     private final List<String> parameters;
+    private final List<Class> parameterTypes;
     private final Block body;
     // this field would be null if we are deserializing from data saved by old version
     private final Class<? extends CpsClosure> closureType;
     private final SourceLocation loc;
 
-    public ClosureBlock(SourceLocation loc, List<String> parameters, Block body, Class<? extends CpsClosure> closureType) {
+    public ClosureBlock(SourceLocation loc, List<Class> parameterTypes, List<String> parameters, Block body, Class<? extends CpsClosure> closureType) {
```
oleg-nenashev: same

kohsuke:
I think @jglick needs to review this.

src/main/java/com/cloudbees/groovy/cps/Builder.java
kohsuke: Not here. It was needed in `ClosureBlock` serialization form.

I'm going to write up the compatibility model of this module, since this is clearly not obvious to many.

src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -556,4 +556,10 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
     void varArgs() {
         assert evalCPS('def fn(String... args) { args.size() }; fn("one string")') == 1;
     }
+
+    @Issue("JENKINS-28277")
+    @Test
+    void currying() {
+        assert evalCPS('def nCopies = { int n, String str -> str*n }; def twice=nCopies.curry(2); twice("foo")') == "foofoo";
```
jglick: Would like to see a test `curry`ing a native method, as well as something in `DefaultGroovyMethods`.

jglick:
Did you check `ncurry`?

jglick:
What about the method pointer operator—already covered somewhere?

src/main/java/com/cloudbees/groovy/cps/impl/CpsClosure.java
```diff
@@ -36,5 +43,15 @@ public Object call(Object arguments) {
         throw new CpsCallableInvocation(def,this,arguments);
     }
 
+    /**
+     * {@link ClosureWriter} generates this function with actual argument types.
+     * Here we approximate by using varargs.
+     * <p>
+     * {@link CurriedClosure} invokes this method directly (via {@link MetaClassImpl#invokeMethod(Class, Object, String, Object[], boolean, boolean)}
+     */
+    public Object doCall(Object... args) {
```
jglick: Having fixed #20, use of varargs here scares me a little.

jglick:
Do not really follow the details, and I think more test coverage is in order, but looks OK so far.

src/main/java/com/cloudbees/groovy/cps/Builder.java
oleg-nenashev: OK, thanks

src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
kohsuke: You can only curry a `Closure`, so I assume you mean test that curries a non-CPS closure, which isn't very interesting IMO but I'm happy to add it. For the same reason, one cannot curry `DefaultGroovyMethods`.

src/main/java/com/cloudbees/groovy/cps/impl/CpsClosure.java
kohsuke: The next step up from here is to have `CpsTransformer` produce AST that defines a new `CpsClosure` subtypes that have the correct type signature in `doCall` method. Not something I want to do unless the need is proven.

PR #22, opened 2016-03-28
Added support for method pointers

kohsuke:
While I'm at #21, might as well.

@reviewbybees  in particular @jglick


jglick:
:bee:

Issue #8, opened 2015-03-13
ArrayList.each method dispatch is buggy


kohsuke:
`Sequence.each` correctly dispatches to `CpsDefaultGroovyMethods.each`, but somehow `ArrayList.each` isn't dispatching to that method. Instead, it finds `DefaultGroovyMethods.each`.

Presumably the way we register `CpsDefaultGroovyMethods` methods into Groovy runtime is incorrect, and the method dispatch cache for `ArrayList` contains stale entry.

To fix this issue, either figure out how to insert meta-method into Groovy, or figure out how to clean up stale cache.

PR #11, opened 2015-03-13
Fix #8: ArrayList.each method dispatch is buggy

jglick:
Fixing #8.

Issue #7, opened 2015-03-13
Generate CpsDefaultGroovyMethods


kohsuke:
It's silly to hand-code `CpsDefaultGroovyMethods` by manually doing CPS transformation.

It'd save me a lot of typing if I can apply `CpsTransformer` during the build process of groovy-cps itself. This way, all I need to do is to cut&paste method definitions from Groovy's `DefaultGroovyMethods` that has `Closure` in the signature.

jglick:
Do not forget to add `DefaultGroovyStaticMethods`, and the things mentioned in `DgmConverter`—I only found a usage of `Closure` in `ProcessGroovyMethods`. None of these methods are likely to be whitelisted or ever used from Pipeline, but we may as well add them just in case.

jglick:
Done in #56.

PR #23, opened 2016-03-31
Handle Object.each { ... }

kohsuke:
Hey @jglick, I delivered on my next airplane project.

See `DGMPatcher` class javadoc for the explanation of the approach. Compare to #11, this handles every subtype of `List`, not just `ArrayList`, no matter what state Groovy is in by the time this code runs.

Once this approach is blessed, we still need to fill in `CpsDefaultGroovyMethods` with all the replacements that we need, and there are hundreds.

@reviewbybees, in particular @jglick

### DIFF
--- a/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
@@ -132,6 +132,11 @@ class DGMPatcher {
         public int hashCode() {
             return Objects.hashCode(declaringClass, name, Arrays.hashCode(nativeParamTypes));
         }
+
+        @Override
+        public String toString() {
+            return declaringClass.getName() + "." + name + Arrays.toString(nativeParamTypes);
+        }
     }
 
     /**
@@ -145,7 +150,10 @@ class DGMPatcher {
      */
     DGMPatcher(List<MetaMethod> methods) {
         for (MetaMethod m : methods) {
-            overrides.put(new Key(m),m);
+            MetaMethod old = overrides.put(new Key(m),m);
+            if (old != null) {
+                throw new IllegalStateException("duplication between " + m + " and " + old);
+            }
         }
     }
 
--- a/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
@@ -1,0 +1,314 @@
+package com.cloudbees.groovy.cps.impl;
+
+import com.cloudbees.groovy.cps.CpsDefaultGroovyMethods;
+import com.google.common.base.Objects;
+import groovy.lang.GroovySystem;
+import groovy.lang.MetaClass;
+import groovy.lang.MetaClassImpl;
+import groovy.lang.MetaClassRegistry;
+import groovy.lang.MetaMethod;
+import org.codehaus.groovy.reflection.CachedClass;
+import org.codehaus.groovy.reflection.CachedMethod;
+import org.codehaus.groovy.reflection.ClassInfo;
+import org.codehaus.groovy.reflection.ClassInfo.ClassInfoSet;
+import org.codehaus.groovy.reflection.GeneratedMetaMethod;
+import org.codehaus.groovy.reflection.ReflectionCache;
+import org.codehaus.groovy.runtime.DefaultGroovyMethods;
+import org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl;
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
```diff
@@ -0,0 +1,314 @@
...
+ * <p>
+ * The only read usage of {@link ClassInfo#newMetaMethods} is in {@link CachedClass#getNewMetaMethods()}, and
+ * this method builds its return value from all the super types. This method is then further used by
+ * {@link MetaClassImpl} when it is instantiated and build its own index.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+class DGMPatcher {
+    // we need to traverse various internal fields of the objects
+    private final Field MetaClassImpl_myNewMetaMethods = field(MetaClassImpl.class,"myNewMetaMethods");
```
jglick: My belly is beginning to hurt… :-)

jglick:
Fixes #8 I suppose.

src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -1,6 +1,8 @@
 package com.cloudbees.groovy.cps
 
+import com.cloudbees.groovy.cps.impl.ContinuationGroup
 import com.cloudbees.groovy.cps.impl.CpsCallableInvocation
+import com.cloudbees.groovy.cps.impl.DGMPatcher
```
jglick: unused imports

src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
```diff
@@ -0,0 +1,314 @@
...
+            this(m.getDeclaringClass().getTheClass(), m.getName(), m.getNativeParameterTypes());
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Key key = (Key) o;
+            return Objects.equal(declaringClass, key.declaringClass) &&
+                    Objects.equal(name, key.name) &&
```
jglick: Note to self: this is the Guava version, not the Java 7 version.

jglick: (Seems we are not running Animal Sniffer on this project. Would be a good idea.)

jglick:
#11 was ugly as sin, and this is uglier than sin, but :bee: assuming it works—I will try to prepare an integration test.

> we still need to fill in `CpsDefaultGroovyMethods` with all the replacements that we need

Best done as a separate PR; tracked as #7.

jglick:
Might be worth making the patch to `evalCPS` mentioned [here](https://github.com/cloudbees/groovy-cps/pull/11#issuecomment-78956805) to demonstrate that the new approach is not vulnerable to cache effects (or at least not this one).

jglick:
:bug: does not seem to work from Pipeline script; see downstream PR.

jglick:
Does not seem to a problem with `CpsDefaultGroovyMethods` itself, since a `println` added to `each` is printed from the unit test but not from the integration test.

My suspicion is that the replacement is happening, but then the original version is getting put back in at some point.

jglick:
Confirmed in a debugger that binary `DefaultGroovyMethods.each(Object, Closure)` is being called on the `['a', 'b', 'c']` list. (There are a lot of earlier calls from the Groovy compiler while parsing the script.)

jglick:
The plot thickens. If you change `true` to `false` to turn off the sandbox, the test still fails—but for a totally different reason: [JENKINS-27421](https://issues.jenkins-ci.org/browse/JENKINS-27421), suggesting that with the sandbox off we _are_ using `CpsDefaultGroovyMethods`, which then explodes because the list iterator is unusable in Pipeline.

jglick:
:bug: until the proper fix location of the two identified bugs has been determined.

jglick:
Working on a JENKINS-27421 fix. Wrote what I thought should work, but is not working yet for reasons TBD.

jglick:
OK, think I have it working now, nothing wrong in this PR, :bee:

kohsuke:
groovy-cps 1.17 released with this change

jglick:
Not only does not work with Groovy 2.x (and thus in Jenkins 2.x), but prevents _anything_ from working. Tracked in [JENKINS-34064](https://issues.jenkins-ci.org/browse/JENKINS-34064).

PR #24, opened 2016-04-06
[JENKINS-34064] Groovy 2.x fixes

jglick:
[JENKINS-34064](https://issues.jenkins-ci.org/browse/JENKINS-34064)

Amends #23. Use with `workflow-cps-plugin`

``` diff
diff --git a/pom.xml b/pom.xml
index 336f94c..88ac635 100644
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
         <dependency>
             <groupId>com.cloudbees</groupId>
             <artifactId>groovy-cps</artifactId>
-            <version>1.7</version>
+            <version>1.8-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.ui</groupId>
```

When running, say,

```
mvn -Dtest=org.jenkinsci.plugins.workflow.SerializationTest\#eachClosure -Djenkins.version=2.0-beta-2 test
```

this still fails (`CpsDefaultGroovyMethods` does not get used), but at least the build runs. Without this patch, basically everything (e.g., `WorkflowTest#demo`) breaks.

Have been trying to traverse the 2.x graph (as of https://github.com/groovy/groovy-core/pull/395), so far without success:

``` diff
diff --git a/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java b/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
index 31fe372..d804943 100644
--- a/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
@@ -20,6 +20,7 @@ import org.codehaus.groovy.runtime.metaclass.NewInstanceMetaMethod;
 import org.codehaus.groovy.util.AbstractConcurrentMapBase;
 import org.codehaus.groovy.util.AbstractConcurrentMapBase.Segment;
 import org.codehaus.groovy.util.FastArray;
+import org.codehaus.groovy.util.ManagedLinkedList;

 import java.lang.reflect.Field;
 import java.util.ArrayList;
@@ -92,6 +93,10 @@ class DGMPatcher {
     private final Field ClassInfo_globalClassSet = field(ClassInfo.class,"globalClassSet");
     private final Field AbstractConcurrentMapBase_segments = field(AbstractConcurrentMapBase.class,"segments");
     private final Field Segment_table = field(Segment.class,"table");
+    // Groovy 2
+    private final Field GlobalClassSet_items = field(ClassInfo.class.getName() + "$GlobalClassSet", "items");
+    private final Field ManagedLinkedList_head = field(ManagedLinkedList.class, "head");
+    private final Field ManagedLinkedList_Element_next = field(ManagedLinkedList.class.getName() + "$Element", "next");

     /**
      * Used to compare two {@link MetaMethod} by their signatures.
@@ -215,6 +220,15 @@ class DGMPatcher {
             patch(ci.getWeakMetaClass());
 //            patch(ci.getCachedClass());
         } else
+        if (o.getClass().getSimpleName().equals("GlobalClassSet")) {
+            patch(o, GlobalClassSet_items);
+        } else
+        if (o instanceof ManagedLinkedList) {
+            patch(o, ManagedLinkedList_head);
+        } else
+        if (o.getClass().getName().equals(ManagedLinkedList.class.getName() + "$Element")) {
+            patch(o, ManagedLinkedList_Element_next);
+        } else
 // doesn't look like we need to visit this
 //        if (o instanceof CachedClass) {
 //            CachedClass cc = (CachedClass) o;
```

@reviewbybees esp. @kohsuke

### DIFF
--- a/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
@@ -20,11 +20,13 @@ import org.codehaus.groovy.runtime.metaclass.NewInstanceMetaMethod;
 import org.codehaus.groovy.util.AbstractConcurrentMapBase;
 import org.codehaus.groovy.util.AbstractConcurrentMapBase.Segment;
 import org.codehaus.groovy.util.FastArray;
+import org.codehaus.groovy.util.ManagedLinkedList;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
@@ -93,6 +95,16 @@ class DGMPatcher {
     private final Field AbstractConcurrentMapBase_segments = field(AbstractConcurrentMapBase.class,"segments");
     private final Field Segment_table = field(Segment.class,"table");
 
+    // in Groovy 1., ClassInfo.globalClassSet returns ClassInfoSet, which extends AbstractConcurrentMap
+    // in Groovy 2.x, ClassInfo.globalClassSet return private GlobalClassSet, which defines
+    // "ManagedLinkedList items" field.
+    private final Class GlobalClassSet = clazz("org.codehaus.groovy.reflection.ClassInfo$GlobalClassSet");
+    private final Field GlobalClassSet_items = field(GlobalClassSet,"items");
+
+//    private final Field ManagedLinkedList_head = field(ManagedLinkedList.class,"head");
+//    private final Class ManagedLinkedList_Element = clazz(ManagedLinkedList.class.getName()+".Element");
+//    private final Field ManagedLinkedList_Element_next = field(ManagedLinkedList_Element,"next");
+//
     /**
      * Used to compare two {@link MetaMethod} by their signatures.
      */
@@ -197,10 +209,21 @@ class DGMPatcher {
             patch(r.getInstanceMethods());
             patch(r.getStaticMethods());
         } else
+        if (isInstance(GlobalClassSet,o)) {
+            patch(o,GlobalClassSet_items);
+        } else
         // TODO this is redundant. Could simply iterate a collection of fields to see if o is assignable to the defining class.
         if (o instanceof AbstractConcurrentMapBase) {
             // discover all ClassInfo in ClassInfoSet via Segment -> table -> ClassInfo
             patch(o, AbstractConcurrentMapBase_segments);
+        } else
+        if (o instanceof ManagedLinkedList) {
+            for (Iterator itr = ((ManagedLinkedList) o).iterator(); itr.hasNext(); ) {
+                Object item = itr.next();
+                if (patch(item)!=item) {
+                    LOGGER.log(Level.FINE, "Can't replace members of ManagedLinkedList",item);
+                }
+            }
         } else
         if (o instanceof Segment) {
             Segment s = (Segment) o;
@@ -285,6 +308,10 @@ class DGMPatcher {
         return o;
     }
 
+    private boolean isInstance(Class t, Object o) {
+        return t!=null && t.isInstance(o);
+    }
+
     /**
      * Patch a field of an object that's not directly accessible.
      */
@@ -302,16 +329,21 @@ class DGMPatcher {
         }
     }
 
-    private Field field(String owner, String field) {
+    private Class clazz(String name) {
         try {
-            return field(Class.forName(owner), field);
+            return Class.forName(name);
         } catch (ClassNotFoundException x) {
-            LOGGER.log(Level.FINE, "no such class {0}", owner);
+            LOGGER.log(Level.FINE, "no such class {0}", name);
             return null;
         }
     }
 
+    private Field field(String owner, String field) {
+        return field(clazz(owner), field);
+    }
+
     private Field field(Class<?> owner, String field) {
+        if (owner==null)    return null;
         try {
             Field f = owner.getDeclaredField(field);
             f.setAccessible(true);
--- a/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
+++ b/src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
@@ -10,7 +10,6 @@ import groovy.lang.MetaMethod;
 import org.codehaus.groovy.reflection.CachedClass;
 import org.codehaus.groovy.reflection.CachedMethod;
 import org.codehaus.groovy.reflection.ClassInfo;
-import org.codehaus.groovy.reflection.ClassInfo.ClassInfoSet;
 import org.codehaus.groovy.reflection.GeneratedMetaMethod;
 import org.codehaus.groovy.reflection.ReflectionCache;
 import org.codehaus.groovy.runtime.DefaultGroovyMethods;
@@ -30,6 +29,8 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Patches Groovy's method dispatch table so that they point to {@link CpsDefaultGroovyMethods} instead of
@@ -42,7 +43,7 @@ import java.util.Set;
  * at high level, this comes down to the following:
  *
  * <ol>
- * <li>{@link ClassInfoSet} holds references to {@link ClassInfo} where one instance exists for each {@link Class}
+ * <li>{@code ClassInfo.ClassInfoSet} (Groovy 1.x) holds references to {@link ClassInfo} where one instance exists for each {@link Class}
  * <li>{@link ClassInfo} holds a reference to {@link MetaClassImpl}
  * <li>{@link MetaClassImpl} holds a whole lot of {@link MetaMethod}s for methods that belong to the class
  * <li>Some of those {@link MetaMethod}s are {@link GeneratedMetaMethod} that points to methods defined on {@link DefaultGroovyMethods}
@@ -79,6 +80,9 @@ import java.util.Set;
  * @author Kohsuke Kawaguchi
  */
 class DGMPatcher {
+
+    private static final Logger LOGGER = Logger.getLogger(DGMPatcher.class.getName());
+
     // we need to traverse various internal fields of the objects
     private final Field MetaClassImpl_myNewMetaMethods = field(MetaClassImpl.class,"myNewMetaMethods");
     private final Field MetaClassImpl_newGroovyMethodsSet = field(MetaClassImpl.class,"newGroovyMethodsSet");
@@ -86,7 +90,7 @@ class DGMPatcher {
     private final Field ClassInfo_dgmMetaMethods = field(ClassInfo.class,"dgmMetaMethods");
     private final Field ClassInfo_newMetaMethods = field(ClassInfo.class,"newMetaMethods");
     private final Field ClassInfo_globalClassSet = field(ClassInfo.class,"globalClassSet");
-    private final Field ClassInfoSet_segments = field(AbstractConcurrentMapBase.class,"segments");
+    private final Field AbstractConcurrentMapBase_segments = field(AbstractConcurrentMapBase.class,"segments");
     private final Field Segment_table = field(Segment.class,"table");
 
     /**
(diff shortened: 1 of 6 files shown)


jglick:
And of course if you update the `groovy` dependency in this POM to 2.4.6, you will get the expected test failure in `eachArray`. (Also `each`, though I am confused since this existed and presumably passed before #23, yet it seems to be testing essentially the same code path…?)

kohsuke:
With these additional changes the serialization test in question now passes for me.

jglick:
Awesome. Next week I will try to pick this up in an integration test.

kohsuke:
To make the `SerializationTest` pass, I also needed https://github.com/jenkinsci/script-security-plugin/pull/71 and https://github.com/jenkinsci/workflow-cps-plugin/commit/a41aa9b0336a6db6b9ee6dc30719a4bb3f117dc9

src/main/java/com/cloudbees/groovy/cps/impl/DGMPatcher.java
```diff
@@ -285,6 +308,10 @@ private Object patch(Object o) {
         return o;
     }
 
+    private boolean isInstance(Class t, Object o) {
```
jglick: Clearer to inline this I think.

jglick:
The fix does not seem to work, either in the unit test here, or in the downstream integration test. The closure is run only once.

jglick:
Hmm, after

```diff
diff --git a/pom.xml b/pom.xml
index f511471..3731571 100644
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
     <dependency>
       <groupId>org.codehaus.groovy</groupId>
       <artifactId>groovy</artifactId>
-      <version>1.8.9</version>
+      <version>2.4.6</version>
       <!--
         let the user of this library select the right flavor of groovy,
         including groovy vs groovy-all
```

I tried

    mvn clean test -Dtest=CpsTransformerTest

which succeeds, yet

    mvn clean test

fails:

```
Running com.cloudbees.groovy.cps.CpsTransformerTest
Tests run: 56, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 4.763 sec <<< FAILURE!
eachArray(com.cloudbees.groovy.cps.CpsTransformerTest)  Time elapsed: 0.029 sec  <<< FAILURE!
Assertion failed: 

assert resultInCps==sh.evaluate(script)
       |          | |  |        |
       11         | |  16        
                  | |                       def x = 10;
                  | |                       [1, 2, 3].each { y -> x+=y; }
                  | |                       return x;
                  | |                   
                  | groovy.lang.GroovyShell@1199fe66
                  false

	at org.codehaus.groovy.runtime.InvokerHelper.assertFailed(InvokerHelper.java:402)
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.assertFailed(ScriptBytecodeAdapter.java:650)
	at com.cloudbees.groovy.cps.AbstractGroovyCpsTest.evalCPS(AbstractGroovyCpsTest.groovy:50)
	at com.cloudbees.groovy.cps.AbstractGroovyCpsTest$evalCPS.callCurrent(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallCurrent(CallSiteArray.java:52)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:154)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:166)
	at com.cloudbees.groovy.cps.CpsTransformerTest.eachArray(CpsTransformerTest.groovy:620)
	at …
```

jglick:
After

    mvn clean install

here (without the Groovy version patch), from https://github.com/jenkinsci/workflow-cps-plugin/pull/14

    mvn -Dtest=SerializationTest\#eachClosure clean test

fails, as does

    mvn -Dtest=SerializationTest\#eachClosure clean test -Djenkins.version=2.0

with

```
java.lang.AssertionError: Started
[Pipeline] node
Running on master in /tmp/junit8330663951182389291/junit2517820064459358411/workspace/demo
[Pipeline] {
[Pipeline] writeFile
[Pipeline] semaphore
[Pipeline] readFile
[Pipeline] echo
a
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS

	at org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep.waitForStart(SemaphoreStep.java:144)
	at org.jenkinsci.plugins.workflow.SerializationTest$9.evaluate(SerializationTest.java:221)
	at …
```

indicating again that the closure was run just once.

jglick:
The difference in behavior comes down to whether `DGMPatcher.patch` is called once per test suite, or once overall. A logger in `CpsDefaultGroovyMethods._each` confirms that it is not consistently called during failing runs.

To see the failure it suffices to run

    mvn test -Dtest=CpsTransformerTest,GreenThreadTest

jglick:
Successful run has:

```
running each(List, Closure) on []
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
running each(List, Closure) on [1, 2, 3]
running each(List, Closure) on [1, 2, 3]
running each(List, Closure) on [0, 1, 2, 3, 4]
running each(List, Closure) on [0, 1, 2, 3, 4]
running each(List, Closure) on [0, 1, 2, 3, 4]
running each(List, Closure) on [0, 1, 2, 3, 4]
running each(List, Closure) on [0, 1, 2, 3, 4]
running each(List, Closure) on [0, 1, 2, 3, 4]
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

whereas a failing run has:

```
running each(Object, Closure) on [org.codehaus.groovy.ast.Parameter@142269f2[name:y type: java.lang.Object, hasDefaultValue: false]]
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
running each(Object, Closure) on [org.codehaus.groovy.ast.Parameter@2d96543c[name:y type: java.lang.Object, hasDefaultValue: false]]
running each(Object, Closure) on [org.codehaus.groovy.ast.Parameter@3a4e343[name:v type: java.lang.Object, hasDefaultValue: false]]
running each(Object, Closure) on [org.codehaus.groovy.ast.Parameter@2cab9998[name:n type: int, hasDefaultValue: false], org.codehaus.groovy.ast.Parameter@2f7a7219[name:str type: java.lang.String, hasDefaultValue: false]]
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
running each(List, Closure) on [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

kohsuke:
I've tried all three ways you mention here in trying to reproduce the failure you are seeing, but to no avail thus far.

kohsuke:
> The difference in behavior comes down to whether DGMPatcher.patch is called once per test suite, or once overall. A logger in CpsDefaultGroovyMethods._each confirms that it is not consistently called during failing runs.

`DGMPatcher.patch()` gets invoked only once through the static initializer of a class, never "called once per test suite." Presumably Jesse means the difference comes down to the state of Groovy runtime when `DGMPatcher.patch()` is called.

jglick:
> I've tried all three ways you mention here in trying to reproduce the failure you are seeing, but to no avail thus far.

Try with Java 8. Passes for me too in 7, but fails in 8.

jglick:
…but passes on 8 with Groovy 2.4.7.

abayer:
Since we've moved to Groovy 2.4.7 in 2.7.x and later lines, is it time to revisit this?

jglick:
You mean, assuming we drop 1.x support in `workflow-cps`. I am OK with that, but it would not help with this. The problem here is that a fix (or hack, really) was developed for the issue that works with Groovy 1, but last I checked no one has managed to develop an analogue that works with Groovy 2. If we could, we would have merged this long ago, with at worst a simple 1-2 runtime switch.

Anyway I am not convinced `DGMPatcher` is the right approach. It is extremely brittle. I suspect the hooks used by `GroovyCategorySupport` would offer a simpler fix.

abayer:
Ok, I'm gonna try to get to understand this a bit.

abayer:
So it looks like https://issues.apache.org/jira/browse/GROOVY-6263 is the reason we can't use `GroovyCategorySupport` currently? That's fixed in 2.5.0-alpha-1, which is in the process of a release vote as we speak. Obviously, we don't want to depend on an alpha/beta/rc of Groovy, but I can start doing testing against that alpha release in the interim. I'm assuming https://github.com/cloudbees/groovy-cps/blob/master/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java#L67-L92 isn't actually the right way to use a category, given that the comment says it would affect the entire JVM...

jglick:
I have no clue what that comment refers to. FWIW I was not suggesting actually using `GroovyCategorySupport`—it depends on `ThreadLocal` so there is no way it would work in CPS-transformed code, which uses virtual threads—but rather to study which (possibly “internal”) calls it makes in its implementation, and try to use those to convince the Groovy method dispatch system to honor our overrides of `DefaultGroovyMethods`. If that is even possible.

abayer:
fwiw, I've started testing with Groovy 2.5.0-alpha-1 and have hit errors in `CpsTransformerTest` - specifically, `each`, `eachArray`, and `superClass`. I'm investigating to determine what changes in Groovy caused these.

abayer:
I tell a lie - the problem change actually happened in 2.4.. With the following code...

```groovy
import org.codehaus.groovy.runtime.InvokerHelper

public class A {
    @Override
    public String toString() {
        return "base"
    }
}

public class B extends A {
    @Override
    public String toString() {
        return "x" + super.toString()
    }
}

MetaClass mc = InvokerHelper.getMetaClass(B.class);

assert 'super$2$toString' == mc.getMethodWithCaching(B.class, "toString", null, true).getName();
```

...it passes with 2.4.7 and fails with 2.4.8. That results in `DefaultInvoker.superCall(Class,Object,String,Object[])` barfing out with a `CpsCallableInvocation` rather than the `super` call, as best as I can tell.

Gonna find the actual change that caused this now. =)

abayer:
And https://github.com/apache/groovy/commit/0a6789d06cc6451fcfee174ba638c0494f2827ef is the culprit commit. Now the question is do we want to open a bug against Groovy for this change from 2.4.7->2.4.8, or do we want to find a way to work around it in `DefaultInvoker.superCall`?

abayer:
And as I keep digging - `CpsTransformerTest.each` and `.eachArray` broke going from Groovy 2.4.8 to 2.4.9, so unsurprisingly it's a different issue there. Doing some experiments to find that problem now.

abayer:
Found the problem commit in Groovy for `.each` - https://github.com/apache/groovy/commit/b18062a9e524512769e370b1e204a6df5a35c03e

abayer:
Got it! `GlobalClassSet` changed to a new collection behind the scenes. I've got fixes for that *and* for the `superClass` failure (though I don't love that fix - it has the same result as the previous behavior but it feels awkward) up at https://github.com/abayer/groovy-cps/tree/jenkins-34064-continued

jglick:
Not quite sure what it is you are fixing here—do you think there are regressions in Groovy 2.4.8 that have not yet been found by users, or what?

abayer:
There definitely is - https://issues.apache.org/jira/browse/GROOVY-8140 is the `superClass` failure, legit regression in Groovy with a PR up at https://github.com/apache/groovy/pull/520 that I'm hoping will make it into 2.5.0-alpha-2. The `each` and `eachArray` failures aren't due to a bug in Groovy, just a change in `GlobalClassSet` internally to Groovy with an easy fix.

abayer:
So fwiw, https://github.com/abayer/groovy-cps/tree/jenkins-34064-continued, building against the latest Groovy 2.5.0-SNAPSHOT, has `CategorySupport` working properly. Woot.

abayer:
I just realized what you meant re `GroovyCategorySupport`, so now I need to figure out what all that would entail when not done thread-wide.

abayer:
Closing in favor of https://github.com/cloudbees/groovy-cps/pull/51

jglick:
BTW @abayer you do not need to close a PR which you propose to subsume in another PR; if the new one is merged, GH will automatically treat this one as merged, too.

Issue #26, opened 2016-06-24
Interfaces cannot be declared


Fiouz:
Declaring an interface breaks CPS compilation, this prevents basic OOP design in scripts

```groovy
            interface Strategy {
                Closure process(Object event)
            }
            return true
```
```
java.lang.NullPointerException: Cannot invoke method visit() on null object
        at org.codehaus.groovy.runtime.NullObject.invokeMethod(NullObject.java:77)
        at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:45)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:42)
        at org.codehaus.groovy.runtime.callsite.NullCallSite.call(NullCallSite.java:32)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:42)
        at org.codehaus.groovy.ast.stmt.BlockStatement$visit.call(Unknown Source)
        at com.cloudbees.groovy.cps.CpsTransformer.visitMethod(CpsTransformer.groovy:166)
        at sun.reflect.GeneratedMethodAccessor12.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:90)
        at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:233)
        at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:361)
        at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:903)
        at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.callCurrent(PogoMetaClassSite.java:66)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:141)
        at com.cloudbees.groovy.cps.CpsTransformer$_call_closure1.doCall(CpsTransformer.groovy:95)
        at sun.reflect.GeneratedMethodAccessor11.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #33, opened 2016-06-27
Do not touch interfaces

kohsuke:
They need no transformation. See issue #26 and PR #27.

This PR incorporates changes from PR #27.

@reviewbybees

### DIFF
--- a/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
+++ b/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
@@ -627,7 +627,6 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
 
     @Issue('https://github.com/cloudbees/groovy-cps/issues/26')
     @Test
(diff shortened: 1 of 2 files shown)


src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -623,4 +624,24 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
...
+
+    @Issue('https://github.com/cloudbees/groovy-cps/issues/26')
+    @Test
+    void interfaceDeclaration() {
+        assert evalCPS('''
+            interface Strategy {
+                Closure process(Object event)
+            }
+            return true
+        ''') == true;
```
jglick: Would be more reassuring to verify that you can actually *use* the newly defined interface somehow. Implement it, define a variable of interface type, assign an instance of the implementing class, call the interface method.

jglick:
:bee:

PR #34, opened 2016-07-13
Fixes #31: Constructor coercion support

### LINKED ISSUE #31: List to constructor coercion not supported
Maybe a consequence of #12 (incomplete type checking), CPS does not support constructor list:

http://groovy-lang.org/semantics.html#_list_and_map_constructors

> In addition to the assignment rules above, if an assignment is deemed invalid, in type checked mode, a list literal or a map literal A can be assigned to a variable of type T if:
>  the assignment is a variable declaration and A is a list literal and T has a constructor whose parameters match the types of the elements in the list literal

``` groovy
        assert evalCPS('''\
            File f = ['/parent', 'name']
            return f
        '''.stripIndent()) == new File('/parent', 'name')
```

The CPS execution return value is actually the `List` instance not being converted to a proper `File`:

```
Assertion failed: 

assert resultInCps==sh.evaluate(script)
       |          | |  |        |
       |          | |  |        File f = ['/parent', 'name']
       |          | |  |        return f
       |          | |  \parent\name
       |          | groovy.lang.GroovyShell@14a002b1
       |          false
       [/parent, name]
```

svanoort:
Fixes #31. Builds on top of https://github.com/cloudbees/groovy-cps/pull/32 by adding support for type-checked list constructors.  Uses the Groovy internal method to do type coercion before storage, and Envs store variable types when the variables are declared.

TODO:
- [x] Enable testcase
- [x] Implementation, including test passes
- [x] Basic test of variable type storage

This should provide the basic implementation and cover main cases.  If there are special cases that need additional work to support, those can be later efforts.

### DIFF
--- a/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
+++ b/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
@@ -707,6 +707,7 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor 
         if (ref instanceof VariableExpression /* local variable */
         ||  ref instanceof Parameter) {
             makeNode("localVariable") {
+                loc(exp)
                 literal(exp.name)
             }
         } else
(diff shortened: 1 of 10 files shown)


kohsuke:
Looking good so far!

svanoort:
@reviewbybees  especially @kohsuke

stephenc:
:bee: AIUI (but probably want feedback from @kohsuke & @jglick before merging)

PR #35, opened 2016-07-14
Documentation take #1

kohsuke:
The session with @svanoort today gave me some ideas about how to approach the documentation of groovy-cps.

This is my first wave.

@reviewbybees

### DIFF
--- a/doc/block-tree.md
+++ b/doc/block-tree.md
@@ -1,0 +1,2 @@
+# How interpreted program is represented
+TBD
--- a/doc/cps-model.md
+++ b/doc/cps-model.md
@@ -1,0 +1,165 @@
+# How Groovy-CPS models CPS in Java
+In [CPS basics](cps-basics.md), we used the Java8 lambda notation to illustrate CPS translation/execution
+of the program as pseudo code.
+ Unfortunately we cannot actually model groovy-cps like that, because in Java we cannot write a function that
+doesn't return. That pseudo code will very quickly results in `StackOverflowError`.
+
+So in groovy-cps, we use `Next`, `Continuation`, and `Block` to model this.
+
+In pseudo-code, a continuation was a lambda of the form `(v)->{}`. In groovy-cps, `Continuation` interface
+represents a continuation. Here, instead of encapsulating the whole remaining program as single lambda,
+it encapsulates just the tiny logic of "receive the return value of a call and decide what to do next",
+then return "what to do next" as a `Next` object
+
+How does `Next` captures "what to do next"? `Next` is a pair of ...
+
+* `Block`, which represents a program (say, a function call)
+* `Continuation`, which represents how to process the value obtained by evaluating `Block` 
+
+Thus the way the execution goes is like:
+
+1. a `Block` starts execution, which (very quickly) returns a `Next`
+2. We get `Block` and `Continuation` from the `Next` previous step produced
+3. We start evaluating this new `Block` by giving it this new `Continuation`, which (very quickly) returns a `Next`
+4. We get `Block` and `Continuation` from the `Next` previous step produced
+3. We start evaluating this new `Block` by giving it this new `Continuation`, which (very quickly) returns a `Next`
+5. ... repeat ...
+
+Thus, the "interpreter" main loop is as follows. In groovy-cps, this code is actually in `Next.run()`:
+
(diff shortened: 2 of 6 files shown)


doc/cps-model.md
```diff
@@ -0,0 +1,165 @@
...
+    Next n = <<program to execute>>
+    while(true) {
+        n = <<evaluate n.block with n.continuation>>
+    }
+    
+Ths main loop is a little bit like CPU instruction cycle. Fetch next instruction from [Program Counter](https://en.wikipedia.org/wiki/Program_counter),
+execute that one instruction, and repeat that forever. Each instruction is atomic.
+
+## Environment
+We need one more thing to run programs, which is "environments". If you mentally picture how Java program
```
svanoort: Might be worth saying environment == scope  (since "scope" may be an immediately recognizable term for many).

doc/cps-model.md
```diff
@@ -0,0 +1,165 @@
...
+
+    Next n = <<program to execute>>
+    
+    // let's run a program for a bit
+    for (int i=0; i<100; i++) {
+        n = <<evaluate n.block with n.continuation and n.env>>
+    }
+    
+    // let's take a break
+    makeSomeTea();
```
svanoort: :tea:

doc/cps-model.md
```diff
@@ -0,0 +1,165 @@
...
+    
+    // modify n.env so that when the program resumes suspend() returns a value we want
+    n.env.setReturnValueFromSuspendTo("Good job!")
+    
+    // resume the program and now it will print "Good job!"
+    while(true) {
+        n = <<evaluate n.block with n.continuation and n.env>>
+    }
+
+In actual groovy-cps, `Object Continuable.suspend(Object)` plays the role of the pseudo "suspend" function above,
```
svanoort: This is immensely helpful.

src/main/java/com/cloudbees/groovy/cps/Env.java
```diff
@@ -1,21 +1,56 @@
...
 import com.cloudbees.groovy.cps.sandbox.Invoker;
 import groovy.lang.Closure;
 
+import javax.annotation.Nonnull;
 import java.io.Serializable;
 import java.util.List;
 
 /**
- * For variable lookup. This is local variables.
+ * Represents an environment in which {@link Block} is evaluated.
```
svanoort: Worth mentioning that this would be just the defined, named variables (since sometime environment can refer to a lot more about the execution environment and internal state, much like with contexts).

src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
```diff
@@ -10,7 +10,15 @@
 import static java.util.Arrays.*;
 
 /**
- * Invocation of {@link CpsCallable}.
+ * When an CPS-interpreted method is invoked, it immediately throws this error
```
svanoort: Much more explicit, grand.

doc/block-tree.md
```diff
@@ -0,0 +1,2 @@
+# How interpreted program is represented
+TBD
```
svanoort: "TBD" here = awesomeness inbound.

svanoort:
Generally very helpful.  I think it's worth referring to higher level programming concepts (scopes, variable binding, expressions) a bit more in preference to lower-level (stack, machine architecture), since it may be more relatable to some -- other than that, this is immensely useful to someone trying to make sense of the system.

doc/cps-model.md
kohsuke: Scope of variables refer to how far the variable is visible, which includes local variables as well as instance fields. Environment contains local variables, but it has other things like return address and catch handlers, so environment!=scope for me.

src/main/java/com/cloudbees/groovy/cps/Env.java
kohsuke: This is indeed a lot more than the defined named variables.

kohsuke:
More effort to come later.

doc/cps-model.md
svanoort: Maybe == is the wrong expression, but saying it's "scope plus things" makes it a bit easier to follow in the next revision.

PR #36, opened 2016-07-22
Mixing trusted & untrusted code in one CPS session

kohsuke:
See `doc/sandbox.md` for the explanation of what this does.

@reviewbybees

### DIFF
--- a/doc/sandbox.md
+++ b/doc/sandbox.md
@@ -1,0 +1,39 @@
+# Sandboxing CPS execution
+In Jenkins, we want to combine this groovy-cps interpreter with [Groovy sandbox](http://groovy-sandbox.kohsuke.org/),
+so that not only the program gets interpreted (the CPS part) but every method call and property access is also subject to
+access control (the sandbox part.)
+
+Furthermore, sometimes it is desirable to mix some untrusted code (say the script users write) with
+some trusted code (for example ones that define API and libraries used by users), and execute the whole thing
+in CPS.
+
+To do this, groovy-cps has some optional components (`SandboxInvoker` and `SandboxCpsTransformer`) that combines
+Groovy sandbox with groovy-cps.
+
+#How to use it
+Use `SandboxCpsTransformer` instead of `CpsTransformer` to translate the source code, then use `SandboxInvoker`
+instead of `DefaultInvoker` during runtime to perform checks.
+
+To mix trusted code and untrusted code, use two `GroovyShell`, one with `SandboxCpsTransformer` and the other with
+`CpsTransformer`. This requires that you use two classloaders. Then execute the whole thing with `SandboxInvoker`.
+In this way, the portion of the code compiled with `CpsTransformer` will run without security check, while the
+other portion compiled with `SandboxCpsTransformer` will be sandboxed.
+
+#How this works
+
+## Tagging call site
+As `CpsTransformer` translates the program, it adds some metadata to each AST node indicating
+whether the code is trusted or not, wherever it calls methods or access properties. The idea
+is that this will designate whether or not this invocation should be checked for security (untrusted),
+or the invocation should be always allowed to happen (trusted.)
+
+The call site tagging mechanism itself is more general, so it can be used for other purposes,
+for example to record where it came from, etc.
+
+## Runtime check
+During the interpretation, method calls, property access, and other such things will go through `Invoker`.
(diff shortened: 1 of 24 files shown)


jglick:
Without a downstream PR it is unclear what this is for. Would it allow JENKINS-34650 to be solved, as per https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/2?

kohsuke:
You are right that this needs some downstream PR before concluding. That said, this change is ready to be looked at.

I was doing this to enable plugins that are written in Groovy that are CPS transformed, in the context of a joint PoC effort with @abayer.

I think JENKINS-34650 is a good user visible feature out of this change. I'll work on it.

abayer:
:bee:

doc/sandbox.md
```diff
@@ -0,0 +1,39 @@
...
+#How this works
+
+## Tagging call site
+As `CpsTransformer` translates the program, it adds some metadata to each AST node indicating
+whether the code is trusted or not, wherever it calls methods or access properties. The idea
+is that this will designate whether or not this invocation should be checked for security (untrusted),
+or the invocation should be always allowed to happen (trusted.)
+
+The call site tagging mechanism itself is more general, so it can be used for other purposes,
+for example to record where it came from, etc.
```
jglick: So it could at some point be used to record a `CodeSource`, for example?

src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
```diff
@@ -214,6 +205,48 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor
...
+            b = new MethodCallExpression(b, "contextualize",
+                    new PropertyExpression(new ClassExpression(ClassHelper.makeCached(tag)), "INSTANCE"));
+        }
+        return b;
+    }
+
+    /**
+     * {@link Trusted} or {@link Untrusted} tag that gets added to call site.
+     *
+     * @see 'doc/sandbox.md'
```
jglick: Invalid syntax AFAIK; you can specify either a member (as per `link`), or an HTML `<a>` element.

src/main/java/com/cloudbees/groovy/cps/impl/CallSiteBlock.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * {@link Block}s that has {@link CallSiteTag}s.
+ *
+ * @author Kohsuke Kawaguchi
+ * @see Invoker#contextualize(CallSiteBlock)
+ */
+public interface CallSiteBlock extends Serializable, Block {
+    /**
+     * Tags assocaited with this call site.
```
jglick: typo

jglick:
:bee:

doc/sandbox.md
kohsuke: Right

src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
kohsuke: James Nord educated me that I have to use double-quotes.

src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
```diff
@@ -240,7 +240,7 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor
     /**
      * {@link Trusted} or {@link Untrusted} tag that gets added to call site.
      *
-     * @see 'doc/sandbox.md'
+     * @see "doc/sandbox.md"
```
jglick: This is not correct either AFAIK.

src/main/java/com/cloudbees/groovy/cps/sandbox/Invoker.java
```diff
@@ -18,6 +19,7 @@
  * @see Env#getInvoker()
  * @see Continuable#Continuable(Script, Env)
  * @see Envs#empty(Invoker)
+ * @see "doc/sandbox.md"
```
jglick: :ant: This is invalid syntax for `@see` AFAIK.

kohsuke: It is. See http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#see

PR #38, opened 2016-08-08
[JENKINS-34064] DGMPatcher is broken in Groovy 2, so do not try to use it at all

jglick:
Until #24 is finished, the trunk versions of this component (including the 1.7 and 1.8 releases) do not work in Groovy 2 and hence Jenkins 2. So basically suppress #23 for now. With this merged and a 1.9 cut, we should be able to release https://github.com/jenkinsci/workflow-cps-plugin/pull/33.

@reviewbybees esp. @kohsuke


christ66:
:bee:

abayer:
:bee:

PR #40, opened 2016-09-17
Annotate the type to indicate that the CPS transformer has visited

kohsuke:
This came up in the context of https://github.com/jenkinsci/workflow-cps-plugin/pull/59


kohsuke:
@reviewbybees

src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
```diff
@@ -111,6 +111,8 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor
         if (classNode.getField(Verifier.__TIMESTAMP)==null)
             classNode.addField(Verifier.__TIMESTAMP,Modifier.STATIC|Modifier.PRIVATE, ClassHelper.long_TYPE,
             new ConstantExpression(0L));
+
+        classNode.addAnnotation(new AnnotationNode(WORKFLOW_TRANSFORMED_TYPE));
```
jglick: This is to do the extra assertion in the downstream PR IIUC?

kohsuke: Correct

PR #41, opened 2016-09-17
Added support for super.foo(...) calls

kohsuke:


kohsuke:
@reviewbybees

src/test/java/Foo.java
```diff
@@ -0,0 +1,23 @@
+/**
+ * @author Kohsuke Kawaguchi
+ */
+public class Foo {
```
jglick: :ant: what is this for? AFAICT it is unused.

PR #44, opened 2016-11-03
Unset fields at the end of call so we do not hold on to large retained sets

jglick:
Complements https://github.com/jenkinsci/workflow-cps-plugin/pull/83.

@reviewbybees


abayer:
:bee:

PR #46, opened 2017-01-30
[JENKINS-39719] Must make CpsTransformer.iota globally unique

jglick:
[JENKINS-39719](https://issues.jenkins-ci.org/browse/JENKINS-39719)

@reviewbybees

### DIFF
--- a/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
+++ b/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
@@ -22,6 +22,7 @@ import org.codehaus.groovy.syntax.Token
 import javax.annotation.Nonnull
 import java.lang.annotation.Annotation
 import java.lang.reflect.Modifier
+import java.util.concurrent.atomic.AtomicLong
 
 import static org.codehaus.groovy.syntax.Types.*
 
@@ -78,7 +79,7 @@ import static org.codehaus.groovy.syntax.Types.*
  * @author Kohsuke Kawaguchi
  */
 class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor {
-    private int iota=0;
+    private static final AtomicLong iota = new AtomicLong();
     private SourceUnit sourceUnit;
     protected ClassNode classNode;
     protected TransformerConfiguration config = new TransformerConfiguration();
@@ -196,7 +197,7 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor 
               }
          */
 
-        def cpsName = "___cps___${iota++}"
+        def cpsName = "___cps___${iota.getAndIncrement()}"
 
         def builderMethod = m.declaringClass.addMethod(cpsName, PRIVATE_STATIC_FINAL, FUNCTION_TYPE, new Parameter[0], new ClassNode[0],
             new BlockStatement([


src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
```diff
@@ -78,7 +79,7 @@ import static org.codehaus.groovy.syntax.Types.*
  * @author Kohsuke Kawaguchi
  */
 class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor {
-    private int iota=0;
+    private static final AtomicLong iota = new AtomicLong();
```
svanoort: I take it we're relying on classloader isolation between scripts to prevent touching this field across completely separate pipeline executions due to being static?

jglick: No, this class is loaded once in a given Jenkins session. There will be at least one _instance_ per `WorkflowRun` (the problem was that it might be more than one).

svanoort: @jglick My bug-sense (like Spidey sense but nerdier) is tingling from that -- at the very least we're mandating synchronized access across all pipelines which is a potential multithreading bottleneck.

jglick: Well compilation is certainly not cheap to begin with. One `AtomicLong` access per method is hardly going to matter.

svanoort: Specifically my concerns are around it being static.

svanoort: @jglick Other possibility: overflowing this value. 1000 methods * 3000 pipeline jobs * 1000 executions > Integer.MAX_VALUE -- or is there something limiting this to prevent it exceeding that limit in reasonable cases?

Obviously that's a reasonable limit per-script, or per-job, but not necessarily globally.

svanoort: 🐛  for making the an internal counter into a global.  There has got to be a better way to do this.

jglick:
> Integer.MAX_VALUE

That is why it is an `AtomicLong`, not an `AtomicInteger`.

> my concerns are around it being static

And what specifically is your concern? Contention for shared counters is measured in millions of operations per second. We will be happy to get a few dozen, given how expensive Groovy compilation can be—partly due to class loading, which is a vastly more significant source of thread contention (filed separately).

If you are looking for optimization targets, there are plenty of low-hanging fruit.

svanoort:
> That is why it is an AtomicLong

Ah, you're right, I misread that.  That concern is less of one then.

@jglick Is it completely impossible to do this without making the counter fully shared?  This feels like a quick-fix attempt that ought to be solvable in a better way than a static field.

jglick:
> Is it completely impossible to do this without making the counter fully shared?

Of course not. You could do some extra bookkeeping within a `CpsThreadGroup` to stash an iota per build, and then communicate that to each `CpsTransformer` you create somehow. Just a lot more complicated.

There is a totally unrelated risk, of clashes within a build but across Jenkins restarts. I am not even sure how to go about reproducing that (since you would need to compile a supertype but then delay compilation of its subtype, perhaps in a `load`ed script) but it is probably worth defending against just in case.

jglick:
> you would need to compile a supertype but then delay compilation of its subtype

Managed to do this in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/30. It is still OK, since everything is recompiled after the restart anyway.

svanoort:
@jglick I'd *strongly* prefer to see the more complex but safer approach here -- because I'm not sure we're anticipating all the ways that this might introduce problems.  However I don't have grounds to block it without a concrete objection in hand. 🐝

jglick:
> more complex but safer approach

It is not obvious to me that a more complex approach _would_ be safer. There are more moving parts to go wrong. For example, if you keep an iota per build, you need to make it persistent (otherwise there is certainly no advantage over this PR), and then you need to consider what happens when a pre-upgrade build is resumed: do you initialize the iota at 1_000_000 or what?

PR #53, opened 2017-05-12
[GROOVY-6263] Removing private methods so we can use categories

jglick:
[GROOVY-6263](https://issues.apache.org/jira/browse/GROOVY-6263) breaks `SandboxCpsTransformer` in Groovy prior to 2.5, i.e., all versions used by current Jenkins releases.

@reviewbybees

### DIFF
--- a/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
+++ b/src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
@@ -103,7 +103,10 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor 
 
 //        copy(source.ast.methods)?.each { visitMethod(it) }
 //        classNode?.declaredConstructors?.each { visitMethod(it) } // can't transform constructor
-        copy(classNode?.methods)?.each { visitMethod(it) }
+        def methods = classNode?.methods
+        if (methods) {
+            new ArrayList(methods).each { visitMethod(it) }
+        }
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
```diff
@@ -126,11 +129,6 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor
         }
     }
 
-    private <T> List<T> copy(List<T> t) {
```
jglick: Tried package access

```groovy
<T> List<T> copy(List<T> t) {
```

but that was treated as a syntax error. Could have made it `protected` but the method was so trivial it seemed clearer to just inline it.

src/main/groovy/com/cloudbees/groovy/cps/CpsTransformer.groovy
```diff
@@ -141,19 +139,10 @@ class CpsTransformer extends CompilationCustomizer implements GroovyCodeVisitor
                 !node.isAbstract();
     }
 
-    private boolean hasAnnotation(MethodNode node, Class<? extends Annotation> a) {
+    boolean hasAnnotation(MethodNode node, Class<? extends Annotation> a) {
         node.annotations.find { it.classNode.name == a.name } != null
     }
 
-    private boolean extendsFromScript(ClassNode c) {
```
jglick: Apparently unused.

PR #52, opened 2017-05-11
[JENKINS-26481] Patch DGM methods with closure parameters

abayer:
[JENKINS-26481](https://issues.jenkins-ci.org/browse/JENKINS-26481)

Subsumes https://github.com/cloudbees/groovy-cps/pull/51 (which itself subsumes https://github.com/cloudbees/groovy-cps/pull/24).

Note that this is currently a work in progress, adding patched methods one by one, but I wanted to get a PR open for watching the changes as we go.

cc @reviewbybees

PR #54, opened 2017-05-12
[JENKINS-26481] Working on supporting CPS-transformed closures passed to DefaultGroovyMethods

jglick:
Pulls the category trick down from https://github.com/jenkinsci/workflow-cps-plugin/pull/124 (and should become upstream of that if all works out).

Implements #7, picking up artifacts from https://github.com/cloudbees/groovy-cps-dgm-builder/pull/4 (ideally this would be automated by Maven but for now it is manual).

Supersedes #52 (though the expanded tests could be sideported).

@reviewbybees if you are crazy


jglick:
Oops, did not mean to merge that; reverted and refiled as #56.

PR #55, opened 2017-05-15
[JENKINS-44280] Test for overloading methods

abayer:
[JENKINS-44280](https://issues.jenkins-ci.org/browse/JENKINS-44280)

See discussion on https://github.com/cloudbees/groovy-cps-dgm-builder/pull/4 for how this came up.

cc @reviewbybees

### DIFF
--- a/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
+++ b/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
@@ -702,15 +702,34 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
     @Test
     void overloadedMethods() {
         assert evalCPS('''
-            public String bar(List<String> l) {
+            public String bar(List l) {
                 return bar((Iterable)l)
             }
 
-            public String bar(Iterable<String> l) {
+            public String bar(Iterable l) {
                 return "iterable"
             }
 
-            List<String> s = ["a", "b"]
+            List s = ["a", "b"]
+
+            return bar(s)
+        ''') == "iterable"
+    }
+
+    @Issue("JENKINS-44280")
+    @NotYetImplemented
+    @Test
+    void overloadedStaticMethods() {
+        assert evalCPS('''
+            public static String bar(List l) {
+                return bar((Iterable)l)
+            }
+
+            public static String bar(Iterable l) {
+                return "iterable"
+            }
+
+            List s = ["a", "b"]
 
             return bar(s)
         ''') == "iterable"
--- a/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
+++ b/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
@@ -697,6 +697,25 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
         ''') == true;
     }
 
+    @Issue("JENKINS-44280")
+    @NotYetImplemented
+    @Test
+    void overloadedMethods() {
+        assert evalCPS('''
+            public String bar(List<String> l) {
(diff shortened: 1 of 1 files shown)


src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -697,6 +697,25 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
         ''') == true;
     }
 
+    @Issue("JENKINS-44280")
+    @NotYetImplemented
+    @Test
+    void overloadedMethods() {
+        assert evalCPS('''
+            public String bar(List<String> l) {
```
jglick: Being defined on the `Script` I suppose.

src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -697,6 +697,25 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
...
+                return bar((Iterable)l)
+            }
+
+            public String bar(Iterable<String> l) {
```
jglick: I suspect that the type parameters are not relevant here.

src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -697,6 +697,25 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
         ''') == true;
     }
 
+    @Issue("JENKINS-44280")
+    @NotYetImplemented
+    @Test
+    void overloadedMethods() {
+        assert evalCPS('''
+            public String bar(List<String> l) {
+                return bar((Iterable)l)
```
jglick: Interestingly in standard Groovy

```groovy
public String m(List<String> x) {"list"}
public String m(Iterable<String> x) {"iterable"}
List<String> x = ["x"]
assert m(x) == "list" // OK
assert m((Iterable) x) == "iterable" // OK
Iterable<String> x2 = x
assert m(x2) == "iterable" // fails! though Java would say OK
```

abayer: Yeah, Groovy has some non-casting magic going on there.

src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
abayer: Agreed.

src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
abayer: Yup. Same behavior whether we put them in a separate class or not.

jglick:
And yeah if you can find a fix for this, that would be great.

BTW even if you patch it here, I am not completely confident it would work in Pipeline script with sandbox mode. _Probably_ it would, assuming all relevant methods are in fact approved, since in that case we just delegate to Groovy’s standard lookup logic. But I suspect that the selection of method to do the sandbox check on will not be correct, since the interceptor just gets a receiver and a method name and an argument array, and whatever Groovy hacks pick the right overload in some (apparently not all! as above) cases is invisible to this code. IOW you might be asked to whitelist one overload even though the other is actually going to be called. See the discussion in https://github.com/kohsuke/groovy-sandbox/issues/7: the interceptor lacks the information about what method Groovy was actually about to call; it has to basically reverse-engineer the Groovy method dispatch system.

abayer:
Okiedokie.

abayer:
@jglick Are we good to merge this?

PR #56, opened 2017-05-16
[JENKINS-26481] Generate CpsDefaultGroovyMethods, invoke using GroovyCategorySupport

jglick:
I accidentally merged #54 and it seems GitHub offers no way to unmerge a PR.

Pulls the category trick down from jenkinsci/workflow-cps-plugin#124 (and should become upstream of that if all works out).

Implements #7, picking up artifacts from cloudbees/groovy-cps-dgm-builder#4 (ideally this would be automated by Maven but for now it is manual).

Supersedes #52 (though the expanded tests could be sideported).

@reviewbybees


abayer:
I was wondering what happened there...

jglick:
Well I do not know how to implement `sort` so my feeling is that if @kohsuke wants to take it on some time, great, else we do not do it. (It would have been easier had the translation tool been written to reuse `CpsTransformer`, treating its Java input as Groovy code except in special cases like new arrays with initializers, but instead it is a separate beast, and quite hard to work on.)

W.r.t. `StringGroovyMethods`, if you feel like writing a test for it now, I should be able to include it pretty easily (if all goes well), but I am guessing it is rarely used so I was not going to hold things up for that.

abayer:
Ok, sounds reasonable on both counts.

jglick:
@reviewbybees done

PR #60, opened 2017-06-09
Use a trusted Builder for translation, except for Closure.call.

abayer:
Eliminates the need to whitelist, well, a lot of things - i.e.,
anything called by a public method in CpsDefaultGroovyMethods. Still
retains the need to whitelist any of those public methods, and still
keeps the Closure.call contents covered by SandboxInvoker when
appropriate.

Tests are downstream in https://github.com/jenkinsci/workflow-cps-plugin/pull/139

@jglick - I can't actually deploy a SNAPSHOT, so if you could do so, I'd appreciate it. =)

cc @reviewbybees

### DIFF
--- a/dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
+++ b/dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
@@ -95,7 +95,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.TreeSet;
 import javax.annotation.Generated;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.Modifier;
@@ -239,16 +238,20 @@ public class Translator {
             });
             typeVars.put(name, typeVar);
         });
+
         JType type = t(e.getReturnType(), typeVars);
         delegating.type(type);
         m.type(type);
 
         List<JVar> delegatingParams = new ArrayList<>();
         List<JVar> params = new ArrayList<>();
+        Map<String,String> paramClasses = new HashMap<>();
+
         e.getParameters().forEach(p -> {
             JType paramType = t(p.asType(), typeVars);
             delegatingParams.add(e.isVarArgs() && p == e.getParameters().get(e.getParameters().size() - 1) ? delegating.varParam(paramType.elementType(), n(p)) : delegating.param(paramType, n(p)));
             params.add(m.param(paramType, n(p)));
+            paramClasses.put(n(p), paramType.erasure().name());
         });
 
         e.getThrownTypes().forEach(ex -> {
@@ -294,7 +297,9 @@ public class Translator {
         }
         delegatingParams.forEach(p -> delegateCall.arg(p));
 
-        JVar $b = m.body().decl($Builder, "b", JExpr._new($Builder).arg(JExpr.invoke("loc").arg(methodName)));
+        JVar untrustedB = m.body().decl($Builder, "untrustedB", JExpr._new($Builder).arg(JExpr.invoke("loc").arg(methodName)));
+        JVar $b = m.body().decl($Builder, "b", untrustedB.invoke("contextualize").arg(codeModel.ref("com.cloudbees.groovy.cps.sandbox.Trusted").staticRef("INSTANCE")));
+
         JInvocation f = JExpr._new($CpsFunction);
 
         // parameter names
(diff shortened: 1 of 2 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/Builder.java
```diff
@@ -41,6 +41,7 @@
 import com.cloudbees.groovy.cps.impl.YieldBlock;
 import com.cloudbees.groovy.cps.sandbox.CallSiteTag;
 import com.cloudbees.groovy.cps.sandbox.Invoker;
+import com.cloudbees.groovy.cps.sandbox.Trusted;
```
jglick: Unused, revert?

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
```diff
@@ -294,7 +297,9 @@ private void translateMethod(final CompilationUnitTree cut, ExecutableElement e,
         }
         delegatingParams.forEach(p -> delegateCall.arg(p));
 
-        JVar $b = m.body().decl($Builder, "b", JExpr._new($Builder).arg(JExpr.invoke("loc").arg(methodName)));
+        JVar untrustedB = m.body().decl($Builder, "untrustedB", JExpr._new($Builder).arg(JExpr.invoke("loc").arg(methodName)));
+        JVar $b = m.body().decl($Builder, "b", untrustedB.invoke("contextualize").arg(codeModel.ref("com.cloudbees.groovy.cps.sandbox.Trusted").staticRef("INSTANCE")));
```
jglick: Wish I could say I understood what this does, but that applies to pretty much everything in this file…anyway yes I think you are on the right track using `Trusted`—I had forgotten it existed. IIRC introduced for the benefit of `workflow-cps-global-lib`.

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
```diff
@@ -332,10 +337,25 @@ public JExpression visitMethodInvocation(MethodInvocationTree mt, Void __) {
...
                     MemberSelectTree mst = (MemberSelectTree) ms;
-                    inv = $b.invoke("functionCall")
+                    String receiver = mst.getExpression().toString();
+                    String methNm = n(mst.getIdentifier());
+
+                    JVar funcBuilder;
+                    // If we're calling anything other than Closure.call, use the trusted builder, which tags the
+                    // relevant block with Trusted, so we don't traverse the entire call stack with SandboxInvoker.
+                    // Otherwise, use the untrusted builder.
+                    if (paramClasses.getOrDefault(receiver, "null").equals("Closure") &&
```
jglick: Wish there were a cleaner way to do this but I am not sure what.

jglick:
FTR how this winds up looking:

```java
private static<T extends Object >Iterator<T> $each__java_util_Iterator__groovy_lang_Closure(Iterator<T> self, Closure closure) {
    Builder untrustedB = new Builder(loc("each"));
    Builder b = untrustedB.contextualize(com.cloudbees.groovy.cps.sandbox.Trusted.INSTANCE);
    CpsFunction f = new CpsFunction(Arrays.asList("self", "closure"), b.block(b.while_(null, b.functionCall(2028, b.localVariable("self"), "hasNext"), b.block(b.declareVariable(2029, Object.class, "arg", b.functionCall(2029, b.localVariable("self"), "next")), untrustedB.functionCall(2030, b.localVariable("closure"), "call", b.localVariable("arg")))), b.return_(b.localVariable("self"))));
    throw new CpsCallableInvocation(f, null, self, closure);
}
```

lib/src/main/java/com/cloudbees/groovy/cps/Builder.java
abayer: Yup, doing so.

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
abayer: It's basically avoiding needing to make alternate versions of every `Builder` method for "do this but automatically `Trusted`" by instead creating a copy of a normal `Builder` (which may or may not be `Trusted`/sandboxed) that *is* guaranteed to be `Trusted` and using that for everything but `Closure.call`.

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
abayer: Yeah, I spent about forty minutes trying to find something better.

jglick:
Consider #62 instead.

abayer:
Closing in favor of #62.

PR #61, opened 2017-06-09
[FIXED JENKINS-38268] Proper lexical scoping for closure variables.

abayer:
[JENKINS-38268](https://issues.jenkins-ci.org/browse/JENKINS-38268)

Previously, any variables that hadn't been explicitly declared (via
`ClosureCallEnv.declareVariable`) in a `Closure` would end up getting
set in the captured `Env`. Which means we'd end up setting things
globally when we really didn't mean to. This changes that to only set
a variable in the captured `Env` if there's no `locals` variable
defined by that name *and* the captured `Env` has a non-null return
for `getVariableType(name)` - i.e., the variable name has already been
declared in the captured `Env`.

Further tests in a `parallel` will be needed in `workflow-cps`.

cc @reviewbybees

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
@@ -3,7 +3,6 @@ package com.cloudbees.groovy.cps.impl;
 import com.cloudbees.groovy.cps.Continuation;
 import com.cloudbees.groovy.cps.Env;
 
-import javax.annotation.Nonnull;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -28,24 +27,23 @@ class ClosureCallEnv extends CallEnv {
         this.captured = captured;
     }
 
-    public void declareVariable(@Nonnull Class type, @Nonnull String name) {
+    public void declareVariable(Class type, String name) {
         locals.put(name,null);
         getTypes().put(name, type);
     }
 
-    public Object getLocalVariable(@Nonnull String name) {
+    public Object getLocalVariable(String name) {
         if (locals.containsKey(name))
             return locals.get(name);
         else
             return captured.getLocalVariable(name);
     }
 
-    public void setLocalVariable(@Nonnull String name, Object value) {
-        if (locals.containsKey(name) || captured.getLocalVariableType(name) == null) {
+    public void setLocalVariable(String name, Object value) {
+        if (locals.containsKey(name))
             locals.put(name, value);
-        } else {
+        else
             captured.setLocalVariable(name, value);
-        }
     }
 
     public Class getLocalVariableType(String name) {
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
@@ -41,10 +41,10 @@ class ClosureCallEnv extends CallEnv {
     }
 
     public void setLocalVariable(@Nonnull String name, Object value) {
-        if (!locals.containsKey(name) && captured.getLocalVariableType(name) != null) {
-            captured.setLocalVariable(name, value);
-        } else {
+        if (locals.containsKey(name) || captured.getLocalVariableType(name) == null) {
             locals.put(name, value);
+        } else {
+            captured.setLocalVariable(name, value);
         }
     }
 
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
@@ -3,6 +3,7 @@ package com.cloudbees.groovy.cps.impl;
 import com.cloudbees.groovy.cps.Continuation;
 import com.cloudbees.groovy.cps.Env;
 
+import javax.annotation.Nonnull;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -27,23 +28,24 @@ class ClosureCallEnv extends CallEnv {
         this.captured = captured;
     }
 
-    public void declareVariable(Class type, String name) {
+    public void declareVariable(@Nonnull Class type, @Nonnull String name) {
         locals.put(name,null);
         getTypes().put(name, type);
     }
 
-    public Object getLocalVariable(String name) {
+    public Object getLocalVariable(@Nonnull String name) {
         if (locals.containsKey(name))
             return locals.get(name);
         else
             return captured.getLocalVariable(name);
     }
 
-    public void setLocalVariable(String name, Object value) {
-        if (locals.containsKey(name))
-            locals.put(name, value);
-        else
+    public void setLocalVariable(@Nonnull String name, Object value) {
+        if (!locals.containsKey(name) && captured.getLocalVariableType(name) != null) {
             captured.setLocalVariable(name, value);
+        } else {
+            locals.put(name, value);
+        }
     }
 
     public Class getLocalVariableType(String name) {
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallable.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallable.java
@@ -45,6 +45,9 @@ public abstract class CpsCallable implements Serializable {
     protected final void assignArguments(List<?> args, Env e) {
         // TODO: var args
         for (int i=0; i< Math.min(args.size(),parameters.size()); i++) {
+            Class argClass = args.get(i) != null ? args.get(i).getClass() : Object.class;
+
+            e.declareVariable(argClass, parameters.get(i));
             e.setLocalVariable(parameters.get(i), args.get(i));
         }
     }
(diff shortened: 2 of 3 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
```diff
@@ -27,23 +28,24 @@ public ClosureCallEnv(Env caller, Continuation returnAddress, SourceLocation loc
...
-    public void setLocalVariable(String name, Object value) {
-        if (locals.containsKey(name))
-            locals.put(name, value);
-        else
+    public void setLocalVariable(@Nonnull String name, Object value) {
+        if (!locals.containsKey(name) && captured.getLocalVariableType(name) != null) {
             captured.setLocalVariable(name, value);
+        } else {
+            locals.put(name, value);
+        }
```
jglick: Maybe more readably (as a diff, and arguably absolutely too):

```java
if (locals.containsKey(name) || captured.getLocalVariableType(name) == null) {
    locals.put(name, value);
} else {
    captured.setLocalVariable(name, value);
}
```

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -777,4 +777,18 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
...
+        assert evalCPS('''
+def a = [id: 'a', count: 0]
+def b = [id: 'b', count: 0]
+
+def toRun = [a, b].collect { thing -> return { thing.count = thing.count + 1 } }
+
+toRun.each { arg -> arg() }
+
+return [a.count, b.count]
+''') == [1, 1]
```
jglick: And previously this returned `[0, 2]` I suppose?

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -777,4 +777,18 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
...
     }
 
+    @Issue("JENKINS-38268")
+    @Test
+    void lexicalScope() {
+        assert evalCPS('''
+def a = [id: 'a', count: 0]
+def b = [id: 'b', count: 0]
+
+def toRun = [a, b].collect { thing -> return { thing.count = thing.count + 1 } }
```
jglick: Is there a simpler variant that would demonstrate the problem without `[Cps]DefaultGroovyMethods`?

abayer: I'll see what I can do.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
abayer: Correct.

abayer:
@jglick `captured.getLocalVariableType(name) == null` is the case when the variable `name` hasn't been defined in that `ClosureCallEnv`'s scope before. So here, that'd be line 787, `thing` - there's no explicit declaration of `thing` (`CpsCallable.assignArguments` calls `Env.setLocalVariable`, and doesn't call `Env.declareVariable`, so, `ClosureCallEnv.locals` doesn't have `thing` added to it) and `thing` doesn't exist in `ClosureCallEnv.captured` (i.e., the parent `Env`), so we added it to `ClosureCallEnv.locals`. Tada!

lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
abayer: *nod*

jacob-keller: I agree, @jglick version is easier to understand.

jglick:
So it sounds like the bug is that `ClosureCallEnv.declareVariable` is not getting called. Perhaps in `CpsTransformer.visitClosureExpression`? Or perhaps `assignArguments` is wrong—but then how could it work in `CpsFunction`?

jacob-keller:
This seems reasonable to me, but I definitely only have a limited understanding of the code here. Essentially, previously, if we didn't explicitly make a variable local it would end up always being "global"? But I do also wonder why it's not being declareVariable called? Hmmm

abayer:
@jglick It's possible that the problem just can't show up unless you're in `Closure`s - and specifically, in some weird `Closure` stuff like either DGM things or `parallel`, because I can't get the test to fail without using DGM stuff!

jglick:
(Maybe it _does not_ work in functions and we just have not noticed yet, because `FunctionCallEnv.setLocalVariable` implicitly defines variables on first use, whereas it should actually be throwing an error for an undefined variable.)

abayer:
Well, fwiw, I tried adding a `e.declareVariable` call in `CpsCallable.assignArguments` (before the `e.setLocalVariable` call) and the tests all pass with that too. So...yeah. Dunno which is More Right.

jglick:
Patching `assignArguments` makes a lot more sense to me. (Ideally we would have a type for the variable, but currently `CpsCallable` takes only names, because for example [this](https://github.com/cloudbees/groovy-cps/blob/63b6aa1aa3b9414b4b2657ffbb5eeea39b21d115/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java#L210) throws out types—so, for now we have only `Object`.)

abayer:
Ok, updating.

jglick:
> I can't get the test to fail without using DGM stuff

Strange, it seems like it should be a pretty general problem. Did you try inlining the basic implementation of, say, `Iterable.each(Closure)`?

abayer:
@jglick I did not. My brain is a bit fried, so I'll revisit this weekend or Monday. =)

jglick:
Also try patching `FunctionCallEnv.setLocalVariable` to choke on an undefined variable and see if anything fails. I suspect this is an illegitimate situation, once `assignArguments` is fixed.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallable.java
```diff
@@ -45,6 +45,9 @@
     protected final void assignArguments(List<?> args, Env e) {
         // TODO: var args
         for (int i=0; i< Math.min(args.size(),parameters.size()); i++) {
+            Class argClass = args.get(i) != null ? args.get(i).getClass() : Object.class;
```
jglick: This is probably wrong. We do not know on a given invocation what the declared type is supposed to be. I think we need to use `Object` in all cases.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallable.java
```diff
@@ -45,6 +45,9 @@
...
+
+            e.declareVariable(argClass, parameters.get(i));
```
jglick: This is perhaps wrong. The variable may be already declared.

Now I am thinking this is wrong. `assignArguments` is called every time the closure is used. We need to call `declareVariable` when the closure is defined.

jacob-keller: So this should be done in the CpsCloserDef maybe?

jglick:
~I am increasingly convinced that there are missing reflective calls to `declareVariable` in `CpsTransformer`. In the case of `visitMethod`, it should have been there, but the effect is invisible since `FunctionCallEnv` quietly tolerates sets to undeclared variables. But then it is `visitClosureExpression` that is the visible problem.~

Or perhaps not. Every call to `CpsClosureDef.invoke` creates a fresh `ClosureCallEnv`. So that would mean that the current patch would be fine if it just passed `Object.class` unconditionally (pending improvements to `CpsTransformer` to actually record the declared type of closure parameters, if even specified—generally `groovy-cps` does little type checking).

abayer:
One last note before I run off for the day - just verified that if you do something like `[0, 1].each { foo = "bar"; println it }; println foo` in normal Groovy, you get 
```
0
1
bar
```
So it does look like arguments are the only case where the lexical scoping is wrong here pre-patch - they're being treated as if they're the same as `foo = "bar"`, when they should be getting treated the same as `def foo = "bar"`.

jacob-keller:
Really glad we finally found the solution to this long standing issue! 👍

kohsuke:
I was about to say we would be better off declaring variable from `CpsCallable.assignArguments` but it looks like you guys got to the same conclusion.

kohsuke:
Actually, thinking some more about this, this alone might not be enough.

If you think about

```
def x = { y = 3 }
x();
```

then I think y needs to be a local variable of a closure, instead of getting added to the captured environment.

kohsuke:
After writing [this test case](https://gist.github.com/kohsuke/c3fa1cc1ef4e8a5f7341d25440c3136a) it looks like I was wrong. `y` is actually set to the owner, and it goes through `Invoker.setProperty()` and never come to `Env.setLocalVariable`

jglick:
Right, in that case `y` is an implicit property of the `Script` (akin to an undeclared field).

jglick:
Ready for you to

* `mvn clean install source:jar deploy:deploy -DaltDeploymentRepository=maven.jenkins-ci.org::default::https://repo.jenkins-ci.org/snapshots/`
* create a downstream PR in `workflow-cps-plugin` demonstrating that there is no regression, and adding a test or two using `parallel` etc., according to reported problems

abayer:
Test case definitely would be worthwhile. I'll add it later today.

@jglick @kohsuke might it make sense to merge #60 before going any further with this one?

jglick:
Actually I feel less nervous about this one.

abayer:
Fair enough. =) Tests to come in `workflow-cps` later today or Monday!

abayer:
Downstream test added at https://github.com/jenkinsci/workflow-cps-plugin/pull/143

PR #62, opened 2017-06-14
Use trusted builder for translation

jglick:
#60 stripped down to its essentials.

How I got here: I was worried that #60 would open a security vulnerability, specifically calls from `CpsDefaultGroovyMethods` (or its delegates) to user-defined methods that did not happen to be some `Closure.call` overload. So I hit on `Iterable.iterator`, which is also called on user-controlled arguments but for which we do not do the check. There was a snag that `IteratorHack` actually ignores _any_ override of this method, but even with that suppressed, the test did not fail the way I expected:

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java b/src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
index 06f7dbe1..1e119c30 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
@@ -87,7 +87,7 @@ public class IteratorHack {
     }
 
     /** Serializable replacement for {@link List#iterator}. */
-    public static <E> Iterator<E> iterator(List<E> list) {
+    public static <E> Iterator<E> iteratorXXX(List<E> list) {
         // TODO !Caller.isAsynchronous(list, "iterator") && !Caller.isAsynchronous(IteratorHack.class, "iterator", list) when used from a Java 5-style for-loop, so not sure how to sidestep this when running in @NonCPS
         return new Itr<>(list);
     }
diff --git a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
index 90b6a1e9..4d6cf7af 100644
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
@@ -30,10 +30,14 @@ import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 import static org.junit.Assert.*;
+import org.junit.ClassRule;
 import org.junit.Test;
+import org.jvnet.hudson.test.BuildWatcher;
 
 public class CpsFlowDefinition2Test extends AbstractCpsFlowTest {
 
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+
     /**
      * I should be able to have DSL call into async step and then bring it to the completion.
      */
@@ -79,4 +83,25 @@ public class CpsFlowDefinition2Test extends AbstractCpsFlowTest {
         jenkins.assertBuildStatus(Result.FAILURE, jenkins.waitForCompletion(r));
         jenkins.assertLogContains("org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance", r);
     }
+
+    @Test
+    public void sandboxInvokerUsedForReal() throws Exception {
+        WorkflowJob job = jenkins.jenkins.createProject(WorkflowJob.class, "p");
+        job.setDefinition(new CpsFlowDefinition("class L extends ArrayList {\n" +
+                "  public Iterator iterator() {\n" +
+                "    Jenkins.instance.systemMessage = 'bwahaha'\n" +
+                "    super.iterator()\n" +
+                "  }\n" +
+                "}\n" +
+                "def l = new L()\n" +
+                "l.add(1)\n" +
+                "l.collectEntries { e ->\n" +
+                "  [(e): e]\n" +
+                "}\n", true));
+
+        WorkflowRun r = job.scheduleBuild2(0).get();
+        assertEquals(null, jenkins.jenkins.getSystemMessage());
+        jenkins.assertBuildStatus(Result.FAILURE, r);
+        jenkins.assertLogContains("org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance", r);
+    }
 }
```

The sandbox check is still applied:

```
[p #1] org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance
[p #1] 	at org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist.rejectStaticMethod(StaticWhitelist.java:192)
[p #1] 	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor$12.reject(SandboxInterceptor.java:330)
[p #1] 	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onGetProperty(SandboxInterceptor.java:363)
[p #1] 	at org.kohsuke.groovy.sandbox.impl.Checker$4.call(Checker.java:241)
[p #1] 	at org.kohsuke.groovy.sandbox.impl.Checker.checkedGetProperty(Checker.java:238)
[p #1] 	at com.cloudbees.groovy.cps.sandbox.SandboxInvoker.getProperty(SandboxInvoker.java:28)
[p #1] 	at com.cloudbees.groovy.cps.impl.PropertyAccessBlock.rawGet(PropertyAccessBlock.java:20)
[p #1] 	at L.iterator(WorkflowScript:3)
[p #1] 	at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.collectEntries(CpsDefaultGroovyMethods:3564)
[p #1] 	at WorkflowScript.run(WorkflowScript:9)
[p #1] 	at …
```

Seems that the sandbox invoker is somehow stuck to the defining closure—or whatever it is—and you do not need to use a different tag when _calling_ that method. It just works.

So proposing this as an alternative, though I would like to get review by @kohsuke before this gets released.

An example generated method for reference:

```java
private static<T extends Object >Iterator<T> $each__java_util_Iterator__groovy_lang_Closure(Iterator<T> self, Closure closure) {
    Builder b = new Builder(loc("each")).contextualize(com.cloudbees.groovy.cps.sandbox.Trusted.INSTANCE);
    CpsFunction f = new CpsFunction(Arrays.asList("self", "closure"), b.block(b.while_(null, b.functionCall(2028, b.localVariable("self"), "hasNext"), b.block(b.declareVariable(2029, Object.class, "arg", b.functionCall(2029, b.localVariable("self"), "next")), b.functionCall(2030, b.localVariable("closure"), "call", b.localVariable("arg")))), b.return_(b.localVariable("self"))));
    throw new CpsCallableInvocation(f, null, self, closure);
}
```

@reviewbybees

### DIFF
--- a/dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
+++ b/dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
@@ -294,7 +294,8 @@ public class Translator {
         }
         delegatingParams.forEach(p -> delegateCall.arg(p));
 
-        JVar $b = m.body().decl($Builder, "b", JExpr._new($Builder).arg(JExpr.invoke("loc").arg(methodName)));
+        JVar $b = m.body().decl($Builder, "b", JExpr._new($Builder).arg(JExpr.invoke("loc").arg(methodName)).
+            invoke("contextualize").arg(codeModel.ref("com.cloudbees.groovy.cps.sandbox.Trusted").staticRef("INSTANCE")));
         JInvocation f = JExpr._new($CpsFunction);
 
         // parameter names


jglick:
@reviewbybees done

kohsuke:
I added a bit more documentation in #63 to shed some more light in this.

`Jenkins.instance.systemMessage = 'bwahaha'` fails because `L.iterator()` method is still compiled with the 'untrusted' tag. So each line is sandbox executed, regardless of who the caller is, even when the caller is `CpsDefaultGroovyMethods`.

With this change, what we need to look at is if any of the transformed `DefaultGroovyMethods` invoke methods that shouldn't be whitelisted on its own, or if it invokes methods on the arguments that it shouldn't have.

Let's look at `min(Iterable<T> self, Closure closure)` as an example

```
    public static <T> T min(Iterable<T> self, @ClosureParams(FirstParam.FirstGenericType.class) Closure closure) {
        int params = closure.getMaximumNumberOfParameters();
        if (params != 1) {
            return min(self, new ClosureComparator<T>(closure));
        }
        boolean first = true;
        T answer = null;
        Object answerValue = null;
        for (T item : self) {
            Object value = closure.call(item);
            if (first) {
                first = false;
                answer = item;
                answerValue = value;
            } else if (ScriptBytecodeAdapter.compareLessThan(value, answerValue)) {
                answer = item;
                answerValue = value;
            }
        }
        return answer;
    }
```

* It calls `closure.getMaximumNumberOfParameters()` first. Previously this call was sanbox checked, now it's not. Since `closure` is an argument supplied by the untrusted caller, in theory this allows somebody to "trick" the system into calling this method. Could there be a risk in that? Probably not, since it's difficult to imagine this method having any side effect.
* It calls `self.iterator()` implicitly in the for-each loop. `self` is a caller supplied variable, so the same question applies. Is  there a risk in allowing this call? Probably not, since we assume iteration is side-effect free.

... and it continues like that.

The next one is the dangerous example, although this one doesn't have the `Closure` parameter so we don't have this issue in this code:
```
    public static void putAt(Object self, String property, Object newValue) {
        InvokerHelper.setProperty(self, property, newValue);
    }
```
This would allow untrusted caller to set any property on any object, so this would be a big no-no.

`DefaultGroovyMethods` have a lot of methods that get translated. I don't think any of us has or can afford to scrutinize each line like this. I did some casual glance at `DefaultGroovyMethods` and they look mostly mundane. So one way to do this is to say "it's _most likely_ OK" and just merge this.

The other way to look at this is that "most likely" is wreckless, and instead whitelist each method sepately.  In the example of the min method, that would be `Closure.getMaximumNumberOfParameters`, `Iterable.iterator`, `ScriptBytecodeAdapter.compareLessThan`, and so on. To do this, we need to find all these methods that `DefaultGroovyMethods` is calling, which can be done for example by looking at the bytecode and determining what methods are being called.

I trust your judgement on which way you go.

abayer:
@kohsuke fwiw, I've got the full list of methods called over at the initial commit on https://github.com/jenkinsci/script-security-plugin/pull/124 - you can find it at https://github.com/jenkinsci/script-security-plugin/pull/124/commits/5187699a0cc2e8c7887e552feceb281a05fe02b6

kohsuke:
@abyer,

I want to be really precise what you mean by "the full list of methods called over." Can you elaborate? For example, one of the lines is `staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods collectEntries java.lang.Object[] groovy.lang.Closure` but this shouldn't be here, as this method shall never be invoked. It should be calling the same method on `CpsDefaultGroovyMethods` (and any method on `CpsDefaultGroovyMethods` should be safe to invoke.)

Earlier in that file, like `method groovy.lang.Closure getMaximumNumberOfParameters` look good, but then I don't know how this list was assembled.

jglick:
> Is there a risk in allowing this call? Probably not, since we assume iteration is side-effect free.

Well, `Iterable.iterator` is already whitelisted. And as my test showed, while you can override it from your script, the body of that override is then subject to sandbox restrictions.

> instead whitelist each method sepately

We _are_ whitelisting the entry points separately—a hand-selected list of `DefaultGroovyMethods` signatures which have also been translated to `CpsDefaultGroovyMethods`. This change just means that methods called _by_ entry points are not checked.

jglick:
> this shouldn't be here, as this method shall never be invoked

Setting aside for the moment the fact that it _will_ be invoked in non-Pipeline contexts, that is not the point—the whitelist check on the corresponding method in `CpsDefaultGroovyMethods` delegates to the `DefaultGroovyMethods` version. See the downstream PR.

abayer:
@kohsuke Yeah, ignore any of the `DefaultGroovyMethods` or `CpsDefaultGroovyMethods` noise. The relevant methods called by the `CpsDefaultGroovyMethods` methods are all the other ones added in that commit. But as @jglick said, we changed to instead just whitelist the public `DefaultGroovyMethods` methods. There may well be some of those that we don't want to whitelist, though. I just went for a blunt force approach of, well, including all of them.

kohsuke:
> This change just means that methods called by entry points are not checked.
We need to be precise here, so I want to understand what you mean by "methods called by entry points."

If you are whitelisting every method like `Closure.getMaximumNumberOfParameters`, `Iterable.iterator`, `ScriptBytecodeAdapter.compareLessThan`, then you are already operating under the mindse of the latter more conservative answer to the question I posed. In that case, there should have been no need of this PR. CpsDGM could be compiled with the untrusted tag, and everything should have worked fine because everything it calls that matters are supposed to be already whitelisted.

So I still feel like we've got some discrepancy here.

PR #64, opened 2017-08-04
Add groovy-sandbox.version POM property

abayer:
I'm working on tooling to do testing of core + groovy-sandbox +
groovy-cps + workflow-cps + script-security + the rest of the main
Pipeline plugins against bleeding-edge versions of each other and
different versions of Groovy. To do this, it'd be a lot easier if the
dependencies between the various things I'm building/testing had their
versions controlled by properties.

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <groovy.version>2.4.7</groovy.version>
+        <groovy-sandbox.version>1.11</groovy-sandbox.version>
     </properties>
 
     <scm>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -17,6 +17,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <groovy.version>2.4.7</groovy.version>
+        <groovy-sandbox.version>1.11</groovy-sandbox.version>
```
abayer: Whoops, yeah, you're right.

jglick:
note: b5b7d4b38279d5be6f3a08b2764bab7f8e745249

PR #66, opened 2017-08-07
[JENKINS-45904] Implicit string conversion for CPS-transformed classes fails

abayer:
[JENKINS-45904](https://issues.jenkins-ci.org/browse/JENKINS-45904)

Initially just a test reproducing the issue, which is that if you give a CPS-transformed class a `toString` override, and then try to implicitly convert an instance of that class to a string, such as in `echo "This will fail to print ${x}"` where `x` is an instance of such a class, you don't even get an `echo` in the logs - a `CpsCallableInvocation` pops up and throws everything off. Will update as I make progress on a fix.

- [x] Test reproducing issue
- [ ] A fix.

cc @reviewbybees


abayer:
Fair point.

PR #68, opened 2017-08-08
Do nothing in CpsCallableInvocation.fillInStackTrace

jglick:
Just wastes time computing a stack trace for a throwable we intend to catch without logging.

(One approach to [JENKINS-31314](https://issues.jenkins-ci.org/browse/JENKINS-31314) might use the stack trace, but even then I think not.)

@reviewbybees


jglick:
@reviewbybees done

PR #69, opened 2017-08-15
[JENKINS-46088] Downstream testing of double-transformation fix

abayer:
[JENKINS-46088](https://issues.jenkins-ci.org/browse/JENKINS-46088)

Downstream of https://github.com/jenkinsci/groovy-sandbox/pull/37

cc @reviewbybees

### DIFF
--- a/lib/src/test/groovy/com/cloudbees/groovy/cps/sandbox/SandboxInvokerTest.groovy
+++ b/lib/src/test/groovy/com/cloudbees/groovy/cps/sandbox/SandboxInvokerTest.groovy
@@ -87,6 +87,40 @@ String.length()
 ''')
     }
 
+    @Issue("JENKINS-46088")
+    @Test
+    void matcherTypeAssignment() {
+        assert "falsefalse" == evalCpsSandbox('''
+    @NonCPS
+    def nonCPSMatcherMethod(String x) {
+       java.util.regex.Matcher m = x =~ /bla/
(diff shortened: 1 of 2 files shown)


lib/src/test/groovy/com/cloudbees/groovy/cps/sandbox/SandboxInvokerTest.groovy
```diff
@@ -87,6 +87,40 @@ String.length()
...
+Script1:___cps___0()
+Script1:___cps___1()
+Script1:___cps___2()
+Script1.super(Script1).setBinding(Binding)
+Script1.nonCPSMatcherMethod(String)
+ScriptBytecodeAdapter:findRegex(String,String)
+Matcher.matches()
+Script1.cpsMatcherMethod(String)
+ScriptBytecodeAdapter:findRegex(String,String)
+Checker:checkedCast(Class,Matcher,Boolean,Boolean,Boolean)
```
jglick: So in CPS mode it is intentional that this is still passed through the invoker?

As mentioned elsewhere, this whole system probably needs to be reworked, I am just not sure of the details.

abayer: In CPS mode, it should just be going through `SandboxInvoker` and not be sandbox-transformed, yes.

PR #70, opened 2017-08-22
[FIXED JENKINS-46358] Add support for StringGroovyMethods w/closures

abayer:
[JENKINS-46358](https://issues.jenkins-ci.org/browse/JENKINS-46358)

Note that this does *not* include `splitEachLine`, due to that
requiring a private untranslated class.

Downstream tests in `workflow-cps` incoming probably tomorrow AM.

cc @reviewbybees


dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
```diff
@@ -333,20 +336,40 @@ public JExpression visitMethodInvocation(MethodInvocationTree mt, Void __) {
...
+                    if (mst.getExpression() instanceof JCIdent &&
+                            !((JCIdent)mst.getExpression()).sym.toString().equals(fqcn) &&
+                            otherTranslated.containsKey(((JCIdent)mst.getExpression()).sym.toString())) {
+                        inv = $b.invoke("functionCall")
+                                .arg(loc(mt))
+                                .arg($b.invoke("constant").arg(
+                                        otherTranslated.get(((JCIdent)mst.getExpression()).sym.toString()).dotclass()))
+                                .arg(n(mst.getIdentifier()));
+
+                    } else {
```
jglick: So I have _no clue_ what this change is about—can you add some comments to the PR diff giving background?

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsStringGroovyMethodsTest.groovy
```diff
@@ -0,0 +1,101 @@
+package com.cloudbees.groovy.cps
+
+import org.junit.Ignore
+import org.junit.Test
+
+
+class CpsStringGroovyMethodsTest extends AbstractGroovyCpsTest {
+    @Test
```
jglick: Not using parameters like in your original test suite?

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
abayer: It's basically analogous to the change below - but to deal with https://github.com/apache/groovy/blob/f0350c4dc4278db1821b3068df7a6a448dd4bd82/src/main/org/codehaus/groovy/runtime/StringGroovyMethods.java#L3344 specifically. And yes, will do.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsStringGroovyMethodsTest.groovy
abayer: Nope, not enough of them for it to be worth it, basically.

PR #71, opened 2017-09-13
[FIXED JENKINS-32213] Automatically make all CPS classes serializable

abayer:
[JENKINS-32213](https://issues.jenkins-ci.org/browse/JENKINS-32213)

cc @reviewbybees

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
@@ -135,8 +135,8 @@ public class CpsTransformer extends CompilationCustomizer implements GroovyCodeV
                 visitNontransformedStatement(statement);
             }
 
-            if (!classNode.declaresInterface(SERIAIZABLE_TYPE)) {
-                classNode.addInterface(SERIAIZABLE_TYPE);
+            if (!classNode.declaresInterface(SERIALIZABLE_TYPE)) {
+                classNode.addInterface(SERIALIZABLE_TYPE);
             }
 
             // groovy puts timestamp of compilation into a class file, causing serialVersionUID to change.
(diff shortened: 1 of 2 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -1224,6 +1230,8 @@ public void visitBytecodeExpression(BytecodeExpression expression) {
 
     private static final ClassNode METHOD_LOCATION_TYPE = ClassHelper.makeCached(MethodLocation.class);
 
+    private static final ClassNode SERIAIZABLE_TYPE = ClassHelper.makeCached(Serializable.class);
```
svanoort: Must SERIAIZE!

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -133,6 +135,10 @@ public void call(SourceUnit source, GeneratorContext context, ClassNode classNod
                 visitNontransformedStatement(statement);
             }
 
+            if (!classNode.declaresInterface(SERIALIZABLE_TYPE)) {
```
jglick: Well, it could be inheriting `Serializable` from a supertype; did you test that? For that matter, does it hurt anything to unconditionally `addInterface(SERIALIZABLE_TYPE)`?

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -814,4 +814,10 @@ def b = new CpsTransformerTest.Base()
 return (b.&toString)() + (String.getClass().&getSimpleName)()
 ''') == "baseClass"
     }
+
+    @Issue("JENKINS-32213")
+    @Test
+    void allClassesSerializable() {
+        evalCPSonly('class C {}; def c = new C(); assert c instanceof Serializable')
```
jglick: Fine but you also need to test classes which explicitly or implicitly implement `Serializable` already.

Also check behavior of local classes defined inside `@NonCPS` methods, if Groovy even “does” that. These should not be touched.

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
abayer: Yup, you're right, that'll work.

abayer:
@jglick FYI, this is updated and I think it addresses your concerns.

PR #72, opened 2017-09-18
[FIXED JENKINS-44027] Support multiple assignment

abayer:
[JENKINS-44027](https://issues.jenkins-ci.org/browse/JENKINS-44027)

Still need to think about whether to move all the declaration logic into `multipleAssignment` as well, or pull some of it of `multipleAssignment` out into `visitDeclarationExpression`.

cc @reviewbybees

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
@@ -763,6 +763,29 @@ public class CpsTransformer extends CompilationCustomizer implements GroovyCodeV
         BINARY_OP_TO_BUILDER_METHOD.put(KEYWORD_IN, "isCase");
     }
 
+    private void multipleAssignment(final Expression parentExpression,
+                                    TupleExpression tuple,
+                                    ListExpression list) {
+        List<Expression> listExpressions = list.getExpressions();
+        List<Expression> tupleExpressions = tuple.getExpressions();
+        if (listExpressions.size() < tupleExpressions.size()) {
+            // TODO: Better error handling.
+            throw new UnsupportedOperationException();
+        }
+        for (int i = 0, tupleExpressionsSize = tupleExpressions.size(); i < tupleExpressionsSize; i++) {
+            final Expression tupleExpression = tupleExpressions.get(i);
+            final Expression listExpression = listExpressions.get(i);
+            makeNode("assign", new Runnable() {
+                @Override
+                public void run() {
+                    loc(parentExpression);
+                    visit(tupleExpression);
+                    visit(listExpression);
+                }
+            });
+        }
+    }
+
     /**
      * @see
      * org.codehaus.groovy.classgen.asm.BinaryExpressionHelper#eval(BinaryExpression)
@@ -771,14 +794,22 @@ public class CpsTransformer extends CompilationCustomizer implements GroovyCodeV
     public void visitBinaryExpression(final BinaryExpression exp) {
         String name = BINARY_OP_TO_BUILDER_METHOD.get(exp.getOperation().getType());
         if (name != null) {
-            makeNode(name, new Runnable() {
-                @Override
-                public void run() {
-                    loc(exp);
-                    visit(exp.getLeftExpression());
-                    visit(exp.getRightExpression());
-                }
-            });
+            if (name.equals("assign") &&
+                    exp.getLeftExpression() instanceof TupleExpression &&
+                    exp.getRightExpression() instanceof ListExpression) {
+                multipleAssignment(exp,
+                        (TupleExpression)exp.getLeftExpression(),
+                        (ListExpression)exp.getRightExpression());
+            } else {
+                makeNode(name, new Runnable() {
+                    @Override
+                    public void run() {
+                        loc(exp);
+                        visit(exp.getLeftExpression());
+                        visit(exp.getRightExpression());
+                    }
+                });
+            }
             return;
         }
 
@@ -1050,25 +1081,23 @@ public class CpsTransformer extends CompilationCustomizer implements GroovyCodeV
             makeNode("sequence", new Runnable() {
                 @Override
                 public void run() {
+                    // TODO: Possibly move this logic into multipleAssignment(...) or replace that with more specific
+                    // logic.
                     for (Expression e : exp.getTupleExpression().getExpressions()) {
                         final VariableExpression v = (VariableExpression) e;
                         makeNode("declareVariable", new Runnable() {
                             @Override
                             public void run() {
-                                loc(exp);
                                 literal(v.getType());
                                 literal(v.getName());
                             }
                         });
                     }
-                    makeNode("assign", new Runnable() {
-                        @Override
-                        public void run() {
-                            loc(exp);
-                            visit(exp.getLeftExpression());
-                            visit(exp.getRightExpression());
-                        }
-                    });
+                    if (exp.getRightExpression() instanceof ListExpression) {
+                        multipleAssignment(exp,
+                                exp.getTupleExpression(),
+                                (ListExpression)exp.getRightExpression());
+                    }
                 }
             });
         } else {
--- a/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
@@ -770,7 +770,7 @@ public class CpsTransformer extends CompilationCustomizer implements GroovyCodeV
 
         for (int i = 0, tupleExpressionsSize = tupleExpressions.size(); i < tupleExpressionsSize; i++) {
             final Expression tupleExpression = tupleExpressions.get(i);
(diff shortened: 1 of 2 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -763,6 +763,29 @@ public void run() {
         BINARY_OP_TO_BUILDER_METHOD.put(KEYWORD_IN, "isCase");
     }
 
+    private void multipleAssignment(final Expression parentExpression,
+                                    TupleExpression tuple,
+                                    ListExpression list) {
+        List<Expression> listExpressions = list.getExpressions();
+        List<Expression> tupleExpressions = tuple.getExpressions();
+        if (listExpressions.size() < tupleExpressions.size()) {
```
jglick: And if it is greater, then what? Is that legal?

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -1050,25 +1081,23 @@ public void visitDeclarationExpression(final DeclarationExpression exp) {
...
                 @Override
                 public void run() {
+                    // TODO: Possibly move this logic into multipleAssignment(...) or replace that with more specific
+                    // logic.
                     for (Expression e : exp.getTupleExpression().getExpressions()) {
                         final VariableExpression v = (VariableExpression) e;
                         makeNode("declareVariable", new Runnable() {
                             @Override
                             public void run() {
-                                loc(exp);
```
jglick: Why?

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -1050,25 +1081,23 @@ public void visitDeclarationExpression(final DeclarationExpression exp) {
...
-                            loc(exp);
-                            visit(exp.getLeftExpression());
-                            visit(exp.getRightExpression());
-                        }
-                    });
+                    if (exp.getRightExpression() instanceof ListExpression) {
+                        multipleAssignment(exp,
+                                exp.getTupleExpression(),
+                                (ListExpression)exp.getRightExpression());
+                    }
```
jglick: And if not?

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
abayer: I'm aping code in Groovy itself - need to think about it more, though, 'cos I may want to error on it.

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
abayer: Because! More specifically, because we want to invoke https://github.com/cloudbees/groovy-cps/blob/master/lib/src/main/java/com/cloudbees/groovy/cps/Builder.java#L211 - the current code would actually never have worked in the first place, since there's only the above and https://github.com/cloudbees/groovy-cps/blob/master/lib/src/main/java/com/cloudbees/groovy/cps/Builder.java#L215. I also don't think the location is particularly relevant for our use here, since we're just gonna assign it shortly afterwards.

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
abayer: Again, I should probably add erroring out here, but I'm pretty sure we'd never get that far, i.e., that it'd be invalid syntax caught before we get to this point in compilation.

svanoort: Needs a testcase to prove that tho

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -814,4 +814,16 @@ def b = new CpsTransformerTest.Base()
 return (b.&toString)() + (String.getClass().&getSimpleName)()
 ''') == "baseClass"
     }
+
+    @Issue("JENKINS-44027")
+    @Test
+    void multipleAssignment() {
```
svanoort: Needs to have some tests for the weirder/error cases I think to ensure we're exercising the code paths properly

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -763,6 +763,29 @@ public void run() {
...
     }
 
+    private void multipleAssignment(final Expression parentExpression,
+                                    TupleExpression tuple,
+                                    ListExpression list) {
+        List<Expression> listExpressions = list.getExpressions();
+        List<Expression> tupleExpressions = tuple.getExpressions();
+        if (listExpressions.size() < tupleExpressions.size()) {
+            // TODO: Better error handling.
+            throw new UnsupportedOperationException();
```
svanoort: Needs to at least give us a reason here -- just a simple "can't do multi-assign when variable counts differ: {0},{1}" etc etc

abayer:
@svanoort Agreed - will bang those out todayish.

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
abayer: *nod*

abayer: Turns out I was wrong about that. `def (a, b) = 4` will error out due to `Integer.getAt(int)` not being a thing, while `def (a, b) = "what"` will result in `a == 'w'` and `b == 'h'`. So I'm rewriting things accordingly.

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
abayer: See new commit.

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -763,6 +763,41 @@ public void run() {
...
+            // def (a, b, c) = [1, 2] is allowed - c will just be null in that scenario.
+            // def (a, b) = [1, 2, 3] is allowed as well - 3 is just discarded.
+            // def (a, b) = 4 will error due to Integer.getAt(int) not being a thing
+            // def (a, b) = "what" is allowed - a will equal 'w', and b will equal 'h'
+            makeNode("assign", new Runnable() {
+                @Override
+                public void run() {
+                    loc(parentExpression);
+                    visit(tupleExpression);
+                    makeNode("array", new Runnable() {
```
abayer: So this might be able to be simpler, but it's my attempt to approximate https://github.com/apache/groovy/blob/3a035b97712f1e59c792a393c2ea4ec3f32a5871/src/main/org/codehaus/groovy/classgen/asm/BinaryExpressionHelper.java#L445-L464

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
jglick: WTF

abayer: Yeah, I know.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -814,4 +814,48 @@ def b = new CpsTransformerTest.Base()
...
+def (a,b) = "what"
+return a + b
+''') == "wh"
+    }
+
+    @Issue("JENKINS-44027")
+    @Test
+    void mismatchedSizeMultipleAssignment() {
+        assert evalCPS('''
+def (a, b) = ['first', 'second', 'third']
```
svanoort: oh man this is legal?

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -814,4 +814,48 @@ def b = new CpsTransformerTest.Base()
...
+  return e instanceof MissingMethodException
+}
+''') == true
+    }
+
+    @Issue("JENKINS-44027")
+    @Test
+    void arrayLikeMultipleAssignment() {
+        assert evalCPS('''
+def (a,b) = "what"
```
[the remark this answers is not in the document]
abayer: Pretty much, yeah. Groovy will take anything at all array-ish and treat it as an array in some circumstances, like multiple assignment.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
abayer: It totally is!

Issue #73, opened 2017-09-21
Memory use: large amounts of memory used for Env content in CallEnv implementations


svanoort:
CallEnv carries a 'types' array around for every instance:

`private Map<String, Class> types = new HashMap<String, Class>();`

The FunctionCallEnv and BlockScopeEnv and ClosureCallEnv all carry around their own independent maps of locals: 

`private final Map<String,Object> locals = new HashMap<String, Object>();`

These maps can in aggregate be memory consuming - we should make an effort to reduce the memory footprint of CPS code by trying to remove duplication where possible and use compact maps where we can (singletonMap, emptyMap, etc).

svanoort:
Sample heap histogram from one particular user:

> 1:       3225619     1883930096  [B
   2:      13087693     1414354160  [C
   3:      26803765     1286580720  java.util.HashMap
   4:       2553146      612125480  [I
   5:       5234200      428849536  [Ljava.util.HashMap$Node;
   6:       6397241      326882112  [Ljava.lang.Object;
   7:       7396770      236696640  java.util.HashMap$Node
   8:       8773568      210565632  java.lang.String
   9:       8227045      197449080  com.cloudbees.groovy.cps.impl.BlockScopeEnv
  10:       1307970      192596624  [S
  11:       4113772      164550880  com.cloudbees.groovy.cps.impl.FunctionCallEnv

Now, obviously it's hard to tell exactly what's doing what, but since each BlockScopeEnv and FunctionCallEnv has 2 HashMaps (types and locals), that adds up to:  24681634 HashMaps out of 26803765.  

Yes, those two classes hold 92% of all of 'em.

svanoort:
7396770 236696640 java.util.HashMap$Node  means... uh, those maps are largely empty 😢

svanoort:
Resolved by released fix.

PR #74, opened 2017-09-21
Reduce memory of CPS code significantly by lazily initializing locals maps

svanoort:
Provides a solution for #73 by lazily initializing the Maps for scoped environments, and starting them out sized for quite small arrays of locals since often blocks will be empty or have few variables. 

Why does this matter?  See the linked issue for jmap histogram of a real user who has quite a bit of memory consumed by HashMaps from these classes (in fact they're the third largest memory consumer there, even before their contents!). 

This should also reduce GC costs of pipeline rather significantly because the number of objects we create will be quite a bit smaller.

Included in https://github.com/jenkinsci/workflow-cps-plugin/pull/176 to run tests using this.

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
@@ -14,7 +14,10 @@ import java.util.Map;
  */
 // TODO: should be package local once all the impls move into this class
 public class BlockScopeEnv extends ProxyEnv {
+    /** To conserve memory, lazily declared using {@link Collections#EMPTY_MAP} until we declare variables, then converted to a (small) {@link HashMap} */
     private Map<String,Object> locals;
+
+    /** To conserve memory, lazily declared using {@link Collections#EMPTY_MAP} until we declare variables, then converted to a (small) {@link HashMap} */
     private Map<String, Class> types;
 
     public BlockScopeEnv(Env parent) {
@@ -24,8 +27,9 @@ public class BlockScopeEnv extends ProxyEnv {
     public BlockScopeEnv(Env parent, int localsSize) {
         super(parent);
         if (localsSize <= 0) {
-            locals = Collections.emptyMap();
-            types = Collections.emptyMap();
+            // Lazily declare using EMPTY_MAP to conserve memory until we actually declare some variables
+            locals = Collections.EMPTY_MAP;
+            types = Collections.EMPTY_MAP;
         } else {
             locals = Maps.newHashMapWithExpectedSize(localsSize);
             types = Maps.newHashMapWithExpectedSize(localsSize);
@@ -54,7 +58,7 @@ public class BlockScopeEnv extends ProxyEnv {
     /** Because might deserialize old version of class with null value for field */
     private Map<String, Class> getTypes() {
         if (types == null) {
-            this.types = Collections.emptyMap();
+            this.types = Collections.EMPTY_MAP;
         }
         return this.types;
     }
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
@@ -2,6 +2,7 @@ package com.cloudbees.groovy.cps.impl;
 
 import com.cloudbees.groovy.cps.Env;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -12,20 +13,34 @@ import java.util.Map;
  */
 // TODO: should be package local once all the impls move into this class
 public class BlockScopeEnv extends ProxyEnv {
-    private final Map<String,Object> locals = new HashMap<String, Object>();
-    private Map<String, Class> types = new HashMap<String, Class>();
+    private Map<String,Object> locals;
+    private Map<String, Class> types;
 
     public BlockScopeEnv(Env parent) {
-        this(parent, 1);
+        this(parent, 0);
     }
 
-    public BlockScopeEnv(Env parent, int size) {
+    public BlockScopeEnv(Env parent, int localsSize) {
         super(parent);
+        if (localsSize <= 0) {
+            locals = Collections.emptyMap();
+            types = Collections.emptyMap();
+        } else {
+            locals = new HashMap<String, Object>(localsSize);
+            types = new HashMap<String, Class>(localsSize);
+        }
     }
 
     public void declareVariable(Class type, String name) {
+        if (locals == Collections.EMPTY_MAP) {
+            this.locals = new HashMap<String, Object>(1);
+        }
         locals.put(name, null);
-        getTypes().put(name, type);
+
+        if (types == null || types == Collections.EMPTY_MAP) {
+            types = new HashMap<String, Class>(1);
+        }
+        types.put(name, type);
     }
 
     public Object getLocalVariable(String name) {
@@ -38,7 +53,7 @@ public class BlockScopeEnv extends ProxyEnv {
     /** Because might deserialize old version of class with null value for field */
     private Map<String, Class> getTypes() {
         if (types == null) {
-            this.types = new HashMap<String, Class>();
+            this.types = Collections.emptyMap();
         }
         return this.types;
     }
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
@@ -1,6 +1,7 @@
 package com.cloudbees.groovy.cps.impl;
 
 import com.cloudbees.groovy.cps.Env;
+import com.google.common.collect.Maps;
 
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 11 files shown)


svanoort:
@reviewbybees

lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
```diff
@@ -12,16 +13,34 @@
...
+            locals = Collections.emptyMap();
+            types = Collections.emptyMap();
+        } else {
+            locals = new HashMap<String, Object>(localsSize);
+            types = new HashMap<String, Class>(localsSize);
+        }
     }
 
     public void declareVariable(Class type, String name) {
+        if (locals == Collections.EMPTY_MAP) {
```
jglick: Awkward, and fragile since you could accidentally try to write to it, resulting in an `UnsupportedOperationException` which would be confusing.

Why not just leave it null unless and until a request is made to store something in it? You need then only do a null check in a few places, deleting `getTypes`.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
```diff
@@ -39,17 +40,34 @@
...
+    }
+
+    public CallEnv(Env caller, Continuation returnAddress, SourceLocation loc, int localsCount) {
         this.caller = caller;
         this.returnAddress = returnAddress;
         this.callSiteLoc = loc;
         this.invoker = caller==null ? Invoker.INSTANCE : caller.getInvoker();
         assert returnAddress!=null;
+        if (localsCount <= 0) {
+            types = Collections.EMPTY_MAP;
```
jglick: ditto

lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
```diff
@@ -22,14 +23,26 @@
...
-        super(caller,returnAddress,loc);
+        this(caller, returnAddress, loc, captured, closure, 0);
+    }
+
+    public ClosureCallEnv(Env caller, Continuation returnAddress, SourceLocation loc, Env captured, CpsClosure closure, int localsSize) {
+        super(caller,returnAddress,loc, localsSize);
         this.closure = closure;
         this.captured = captured;
+        if (localsSize <= 0) {
+            locals = Collections.emptyMap();
```
jglick: ditto

lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
```diff
@@ -12,16 +13,34 @@
...
+        this(parent, 0);
+    }
+
+    public BlockScopeEnv(Env parent, int localsSize) {
         super(parent);
+        if (localsSize <= 0) {
+            locals = Collections.emptyMap();
+            types = Collections.emptyMap();
+        } else {
+            locals = new HashMap<String, Object>(localsSize);
```
jglick: I think you are not understanding `initialCapacity`. If you intended `localsSize` to be the expected number of entries, and you want to avoid rehashing, then you will need to use something like `localsSize * 4 / 3 + 1`.

lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
svanoort: It's a private field, exposed only by this trivial set of setters, how on early would I "accidentally" try to write to it?

lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
svanoort: Ditto.

lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
svanoort: Ditto.

svanoort:
@abayer Valid point, deployed as version 1.20-20170921.213551-1 and PR pending to demonstrate tests passing

michaelneale:
impressive

lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
```diff
@@ -26,19 +27,19 @@ public BlockScopeEnv(Env parent, int localsSize) {
             locals = Collections.emptyMap();
             types = Collections.emptyMap();
         } else {
-            locals = new HashMap<String, Object>(localsSize);
-            types = new HashMap<String, Class>(localsSize);
+            locals = Maps.newHashMapWithExpectedSize(localsSize);
```
jglick: Ah, handy, did not know about this one.

lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
jglick: Just a stylistic suggestion to make the code more legible (IMO) to the next person to edit this code, perhaps a future version of yourself. At least a comment explaining that this non-null collection field is sometimes mutable and sometimes not would be in order.

The typical idiom is something like

```java
/** Null unless and until something needs to be stored in it. */
@CheckForNull private Map<String,Object> stuff;
@CheckForNull public Object getStuff(String key) {
    return stuff != null ? stuff.get(key) : null;
}
public void putStuff(String key, Object value) {
    if (stuff == null) {
        stuff = new HashMap<>();
    }
    stuff.put(key, value);
}
```

and as noted you can then get rid of `getTypes` and special deserialization logic, because there is no difference between a newly constructed object with a null field, and a deserialized old object with a null field.

BTW if there are some subclasses for which you can know for sure that the collection is always empty, or always contains a single element, then you could optimize further by removing the field from an `abstract` superclass and making the getter-like method `abstract`. One implementation could keep a lazily-initialized `HashMap` field to support any number of entries; another could keep single `final` fields for the key and value. Works best if the setter-like methods can be pushed down into certain concrete subclasses, perhaps narrowing the declared types used by some callers to reflect statically known information about the particular implementation in use. Not sure offhand if that applies here—would need to study the code.

lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
```diff
@@ -12,16 +14,34 @@
...
     }
 
     public void declareVariable(Class type, String name) {
+        if (locals == Collections.EMPTY_MAP) {
+            this.locals = new HashMap<String, Object>(2);
+        }
         locals.put(name, null);
-        getTypes().put(name, type);
+
+        if (types == null || types == Collections.EMPTY_MAP) {
```
jglick: Beware that the Java Platform does not guarantee that `Collections.emptyMap() == Collections.EMPTY_MAP`, though this is true in Oracle JRE 8 at least. The Javadoc says only

> **Implementation Note:** Implementations of this method need not create a separate `Map` object for each call. Using this method is likely to have comparable cost to using the like-named field. (Unlike this method, the field does not provide type safety.)

svanoort:
Pending @abayer to return and sprinkle the re-bee magic on it

svanoort:
Redeployed final version as: 1.20-20170922.134829-2

lib/src/main/java/com/cloudbees/groovy/cps/impl/BlockScopeEnv.java
```diff
@@ -24,8 +27,9 @@ public BlockScopeEnv(Env parent) {
     public BlockScopeEnv(Env parent, int localsSize) {
         super(parent);
         if (localsSize <= 0) {
-            locals = Collections.emptyMap();
-            types = Collections.emptyMap();
+            // Lazily declare using EMPTY_MAP to conserve memory until we actually declare some variables
+            locals = Collections.EMPTY_MAP;
```
jglick: If you wanted, you could avoid an `unchecked`/`rawtypes` warning by going back to `emptyMap()` but switching the test from

```java
locals == Collections.EMPTY_MAP
```

(theoretically unsafe) to

```java
!(locals instanceof HashMap)
```

(definitely safe).

lib/src/main/java/com/cloudbees/groovy/cps/impl/ClosureCallEnv.java
```diff
@@ -14,6 +14,7 @@
  * @author Kohsuke Kawaguchi
  */
 class ClosureCallEnv extends CallEnv {
+    /** Lazily declared using {@link Collections#EMPTY_MAP} until we declare variables, then converted to a (small) {@link HashMap} *//** To conserve memory, lazily declared using {@link Collections#EMPTY_MAP} until we declare variables, then converted to a (small) {@link HashMap} */
```
jglick: typo I suppose

lib/src/main/java/com/cloudbees/groovy/cps/impl/FunctionCallEnv.java
```diff
@@ -12,7 +13,7 @@
  */
 // TODO: should be package local once all the impls move into this class
 public class FunctionCallEnv extends CallEnv {
-    // TODO: delegate?
```
jglick: Did you think about this comment by @kohsuke and decide it is no longer applicable and thus should be deleted?

svanoort: @jglick Yes, at least that was my assessment (I'd looked at using delegation in some of these cases and had to discard it).  

Would be curious what your read of this fairly-cryptic one-word comment is though, maybe you are interpreting it differently than I.

PR #75, opened 2017-09-22
[FIXED JENKINS-47071] Replace BooleanClosureWrapper with serializable

abayer:
[JENKINS-47071](https://issues.jenkins-ci.org/browse/JENKINS-47071)

BooleanClosureWrapper isn't serializable, so attempts to serialize it
(i.e., saveProgram) result in errors. This replaces
BooleanClosureWrapper with CpsBooleanClosureWrapper, which *is*
serializable.

No point in testing here, but actual test coming downstream.

cc @reviewbybees

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsBooleanClosureWrapper.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsBooleanClosureWrapper.java
@@ -1,0 +1,43 @@
+package com.cloudbees.groovy.cps.impl;
+
+import groovy.lang.Closure;
+import org.codehaus.groovy.runtime.callsite.BooleanReturningMethodInvoker;
+
+import java.io.Serializable;
+import java.util.Map;
+
+/**
+ * A serializable equivalent of {@link org.codehaus.groovy.runtime.callsite.BooleanClosureWrapper}, where the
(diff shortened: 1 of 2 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsBooleanClosureWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Bridge for a call based on a map entry. If the call is done on a {@link Closure}
+     * taking one argument, then we give in the {@link Map.Entry}, otherwise we will
+     * give in the key and value.
+     */
+    public <K,V> boolean callForMap(Map.Entry<K, V> entry) {
+        if (numberOfArguments==2) {
+            return call(entry.getKey(), entry.getValue());
+        } else {
+            return call(entry);
```
jglick: Maybe just create a new `BooleanClosureWraper` and delegate to it, so we have less code here?

abayer: Good point.

abayer:
@jglick I have pretty much no idea what you mean in re `PickleFactory`. =)

svanoort:
I have honestly no idea about how this works, so I probably shouldn't be reviewing it :-/

Issue #76, opened 2017-10-09
No way to limit recursion: with bugs, GroovyCPS will consume all memory


svanoort:
There's no way to limit the depth of recursive calls, it just consumes ever-increasing memory as the depth grows.  

The automatic conversion of field access to getter calls means that the following common lazy-loading pattern translates to an unbounded recursive call that will bring down a Jenkins master:

```
def something = null
def getSomething() {
  if (something == null) {
    return "someComplexComputationResult;"
  }
}
def stuff = getSomething();
```

We need a way to limit the depth of the recursion to prevent this scenario and handle appropriately (error out or at least pause execution and force the issue to be addressed).

svanoort:
Consider adding listener hooks that can collect stats when we create/run blocks about how many we're running and how deeply they nest (or use the Envs for that).

jglick:
> automatic conversion of field access to getter calls

This is supposed to be conditional—a field reference inside the same class is _not_ supposed to be translated to a getter. I have some tests proving that this works in various cases (previous critical bugs). If you have found some cases where that does not work, file JIRAs for them and fix them separately.

Of course there can still be true endless recursions in user code not caused by mistranslations.

svanoort:
@jglick I'm happy to look into that as a second PR (it seemed wonky to me, but wasn't sure if it was just groovy being groovy), but as you note this will also protect against both that and user-induced endless recursions. 

Seems reasonable to do this, since it is ready to go and this issue is causing production outages, then work on fixing the field reference problem, no?  That gives us a defense in depth against issues.

jglick:
I am just not sure which is the riskier change at this point.

svanoort:
@jglick Well let me ask this way: do you see anything in this change that looks immediately problematic?  I an agree to hold off releasing this fully until we can do the second thing, but would like to be able to at least hand a SNAPSHOT-based workflow-cps build containing this to people whose production masters are impacted, as an interim measure.

jglick:
I think I found the flaw in the code sample BTW (alluded to in the PR):

```groovy
def something = null
```

defines a local variable, _not_ a field, and Groovy methods are not closures (they cannot access lexically enclosing local variables). I bet

``` groovy
String something = null
```

or

``` groovy
@Field something = null
```

or even perhaps

``` groovy
something = null
```

would make it work as intended.

svanoort:
Released and resolved now.

PR #77, opened 2017-10-09
Protect versus unbounded recursion

svanoort:
Solves #76 albeit in a rather crude way. 

I'd like to do this ultimately in a way that provides better and deeper diagnostic info (i.e. trying to resolve library names and a deeper stack trace within the CPS code), but this is the simplest thing that will resolve it here and now and help point to where to fix the code issue.

Snapshot deployed as: 1.21-20171018.220523-1

# Screenshot of the result when workflow-cps is run with this plugin version in an endlessly recursive situation
screen shot 2017-10-09 at 3 08 42 pm

### DIFF
--- a/dgm-builder/pom.xml
+++ b/dgm-builder/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>com.cloudbees</groupId>
         <artifactId>groovy-cps-parent</artifactId>
-        <version>1.21-stop-recursion-SNAPSHOT</version>
+        <version>1.21-stop-recursion2-SNAPSHOT</version>
     </parent>
 
     <artifactId>groovy-cps-dgm-builder</artifactId>
--- a/dgm-builder/pom.xml
+++ b/dgm-builder/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>com.cloudbees</groupId>
         <artifactId>groovy-cps-parent</artifactId>
-        <version>1.21-stop-recursion2-SNAPSHOT</version>
+        <version>1.21-SNAPSHOT</version>
     </parent>
 
     <artifactId>groovy-cps-dgm-builder</artifactId>
--- a/dgm-builder/pom.xml
+++ b/dgm-builder/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>com.cloudbees</groupId>
         <artifactId>groovy-cps-parent</artifactId>
-        <version>1.21-SNAPSHOT</version>
+        <version>1.21-stop-recursion-SNAPSHOT</version>
     </parent>
 
     <artifactId>groovy-cps-dgm-builder</artifactId>
--- a/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
@@ -12,6 +12,7 @@ import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -239,8 +240,9 @@ public class CpsTransformer extends CompilationCustomizer implements GroovyCodeV
                 new StaticMethodCallExpression(m.getDeclaringClass(), cpsName, new TupleExpression()));
 //                new ConstructorCallExpression(FUNCTION_TYPE, new TupleExpression(params, body)));
 
-        List<Expression> paramExpressions = new ArrayList<>();
-        for (Parameter p : m.getParameters()) {
+        Parameter[] pms = m.getParameters();
+        List<Expression> paramExpressions = new ArrayList<>(pms.length);
+        for (Parameter p : pms) {
             paramExpressions.add(new VariableExpression(p));
         }
         ArrayExpression paramArray = new ArrayExpression(ClassHelper.OBJECT_TYPE, paramExpressions);
@@ -871,21 +873,28 @@ public class CpsTransformer extends CompilationCustomizer implements GroovyCodeV
             public void run() {
                 loc(exp);
 
-                ListExpression types = new ListExpression();
-                ListExpression params = new ListExpression();
+                ListExpression types;
+                ListExpression params;
 
                 // the interpretation of the 'parameters' is messed up. According to ClosureWriter,
                 // when the user explicitly defines no parameter "{ -> foo() }" then this is null,
                 // when the user doesn't define any parameter explicitly { foo() }, then this is empty,
                 if (exp.getParameters() == null) {
+                    types = new ListExpression(Collections.EMPTY_LIST);
+                    params = new ListExpression(Collections.EMPTY_LIST);
                 } else if (exp.getParameters().length == 0) {
-                    types.addExpression(new ClassExpression(OBJECT_TYPE));
-                    params.addExpression(new ConstantExpression("it"));
+                    types = new ListExpression(Collections.<Expression>singletonList(new ClassExpression(OBJECT_TYPE)));
+                    params = new ListExpression(Collections.<Expression>singletonList(new ConstantExpression("it")));
                 } else {
-                    for (Parameter p : exp.getParameters()) {
-                        types.addExpression(new ClassExpression(p.getType()));
-                        params.addExpression(new ConstantExpression(p.getName()));
+                    Parameter[] paramArray = exp.getParameters();
+                    List<Expression> typesList = new ArrayList<Expression>(paramArray.length);
+                    List<Expression> paramsList = new ArrayList<Expression>(paramArray.length);
+                    for (Parameter p : paramArray) {
+                        typesList.add(new ClassExpression(p.getType()));
+                        paramsList.add(new ConstantExpression(p.getName()));
                     }
+                    types = new ListExpression(typesList);
+                    params = new ListExpression(paramsList);
                 }
                 parent.call(types);
                 parent.call(params);
--- a/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
@@ -6,7 +6,7 @@ package com.cloudbees.groovy.cps;
 public interface DepthTrackingEnv extends Env {
 
     /** Limit on how deeply environments can recurse */
-    int MAX_LEGAL_DEPTH = 4000;
+    int MAX_LEGAL_DEPTH = 3000;
 
     public int getDepth();
 }
--- a/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
@@ -1,0 +1,12 @@
+package com.cloudbees.groovy.cps;
+
+/**
+ * @author Sam Van Oort
+ */
+public interface DepthTrackingEnv extends Env {
+
+    /** Limit on how deeply environments can recurse */
+    int MAX_LEGAL_DEPTH = 2000;
+
+    public int getDepth();
+}
--- a/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
@@ -5,8 +5,10 @@ package com.cloudbees.groovy.cps;
  */
 public interface DepthTrackingEnv extends Env {
 
-    /** Limit on how deeply environments can recurse */
-    int MAX_LEGAL_DEPTH = 3000;
+    /** Limit on how deeply environments can recurse.
+     *  Capped somewhat low to try to limit the ability to run a program that will generate a StackOverflowError when serialized. */
+    int MAX_LEGAL_DEPTH = 1024;
 
+    /** Return how deep this environment is within nested closure/function calls. */
     public int getDepth();
 }
--- a/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
@@ -6,7 +6,7 @@ package com.cloudbees.groovy.cps;
 public interface DepthTrackingEnv extends Env {
 
     /** Limit on how deeply environments can recurse */
-    int MAX_LEGAL_DEPTH = 2000;
+    int MAX_LEGAL_DEPTH = 4000;
 
     public int getDepth();
 }
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
@@ -1,6 +1,7 @@
 package com.cloudbees.groovy.cps.impl;
 
 import com.cloudbees.groovy.cps.Continuation;
+import com.cloudbees.groovy.cps.DepthTrackingEnv;
 import com.cloudbees.groovy.cps.Env;
 import com.cloudbees.groovy.cps.Next;
 import com.cloudbees.groovy.cps.sandbox.Invoker;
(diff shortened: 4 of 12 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
```diff
@@ -57,6 +60,11 @@ public CallEnv(Env caller, Continuation returnAddress, SourceLocation loc, int l
         } else {
             types = Maps.newHashMapWithExpectedSize(localsCount);
         }
+        depth = (caller instanceof  DepthTrackingEnv) ? depth = ((DepthTrackingEnv) caller).getDepth() + 1 : 1;
+
+        if (depth > DepthTrackingEnv.MAX_LEGAL_DEPTH) {
+            throw new StackOverflowError("Excessively nested blocks/function calls exceed limit of"+DepthTrackingEnv.MAX_LEGAL_DEPTH+":"+depth);
```
alexanderrtaylor: need a space after "of". But do you think there is a way in here to point to the user's code?

svanoort: Yes, I have a commit locally that fixes the space issue (pushing it publicly now) and I'm working on how to point the user's code at it

svanoort:
@reviewbybees

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -239,8 +240,9 @@ CpsFunction ___cps___N() {
                 new StaticMethodCallExpression(m.getDeclaringClass(), cpsName, new TupleExpression()));
 //                new ConstructorCallExpression(FUNCTION_TYPE, new TupleExpression(params, body)));
 
-        List<Expression> paramExpressions = new ArrayList<>();
-        for (Parameter p : m.getParameters()) {
+        Parameter[] pms = m.getParameters();
+        List<Expression> paramExpressions = new ArrayList<>(pms.length);
+        for (Parameter p : pms) {
```
jglick: It looks like all the changes to this file are some unrelated optimization. If so, would better be filed as a separate PR.

lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
```diff
@@ -0,0 +1,12 @@
+package com.cloudbees.groovy.cps;
+
+/**
+ * @author Sam Van Oort
+ */
+public interface DepthTrackingEnv extends Env {
+
+    /** Limit on how deeply environments can recurse */
+    int MAX_LEGAL_DEPTH = 3000;
```
jglick: BTW how does this compare to default stack size for a typical JVM and effective recursion depth in that case?

lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public int getDepth();
```
jglick: Javadoc please—the return value here is by no means self-explanatory.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
```diff
@@ -38,6 +39,8 @@
 
     private Invoker invoker;
 
+    int depth;
```
jglick: After deserialization of a program running prior to the upgrade, this will be 0. Is that expected and safe?

lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
```diff
@@ -57,6 +60,15 @@ public CallEnv(Env caller, Continuation returnAddress, SourceLocation loc, int l
         } else {
             types = Maps.newHashMapWithExpectedSize(localsCount);
         }
+        depth = (caller instanceof  DepthTrackingEnv) ? depth = ((DepthTrackingEnv) caller).getDepth() + 1 : 1;
```
jglick: Using return value of assignment, confusingly and probably incorrectly.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
```diff
@@ -57,6 +60,15 @@ public CallEnv(Env caller, Continuation returnAddress, SourceLocation loc, int l
...
+
+        if (depth > DepthTrackingEnv.MAX_LEGAL_DEPTH) {
+            if (loc != null) {
+                throw new StackOverflowError("Excessively nested closures/functions at "+loc.toString()+" - look for unbounded recursion - call depth: "+depth);
```
jglick: BTW no need for `.toString()` here.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
```diff
@@ -57,6 +60,15 @@ public CallEnv(Env caller, Continuation returnAddress, SourceLocation loc, int l
...
+            } else {
+                throw new StackOverflowError("Excessively nested blocks/function - look for unbounded recursion - call depth: "+depth);
```
jglick: Blocks here but closures above? Intentional?

lib/src/main/java/com/cloudbees/groovy/cps/impl/ProxyEnv.java
```diff
@@ -13,11 +14,14 @@
  *
  * @author Kohsuke Kawaguchi
  */
-public class ProxyEnv implements Env {
+public class ProxyEnv implements DepthTrackingEnv {
     protected final Env parent;
 
+    int depth = 0;
```
jglick: Unused initializer—overwritten by constructor.

lib/src/test/groovy/com/cloudbees/groovy/cps/impl/FunctionCallBlockTest.groovy
```diff
@@ -21,6 +22,28 @@ class FunctionCallBlockTest extends AbstractGroovyCpsTest {
...
+            assert evalCPSonly("""
+                def thing = null;
+                def getThing() { 
+                    return thing == null;
+                }
+                def stuff = getThing();
+            """, 25000) == "cheese";
+        } catch (StackOverflowError soe) {
+            println "PASSED: expected exception thrown for endlessly recursive function, see trace below";
+            soe.printStackTrace()
```
jglick: Maybe assert that the detail message has some kind of useful diagnostic.

lib/src/test/groovy/com/cloudbees/groovy/cps/impl/FunctionCallBlockTest.groovy
```diff
@@ -21,6 +22,28 @@ class FunctionCallBlockTest extends AbstractGroovyCpsTest {
...
+    void infiniteRecursion() {
+        try {
+            assert evalCPSonly("""
+                def thing = null;
+                def getThing() { 
+                    return thing == null;
+                }
+                def stuff = getThing();
+            """, 25000) == "cheese";
+        } catch (StackOverflowError soe) {
```
jglick: You need to `fail` if the error is _not_ thrown.

lib/src/test/groovy/com/cloudbees/groovy/cps/impl/FunctionCallBlockTest.groovy
```diff
@@ -21,6 +22,28 @@ class FunctionCallBlockTest extends AbstractGroovyCpsTest {
...
+
+    @Test
+    void infiniteRecursion() {
+        try {
+            assert evalCPSonly("""
+                def thing = null;
+                def getThing() { 
+                    return thing == null;
+                }
+                def stuff = getThing();
```
jglick: Cf. the associated issue. It seems this _is_ a buggy program (reproducible using command-line Groovy—`def thing` does not declare a field), but it might be clearer for test purposes to use a self-evidently endless recursion, such as:

```groovy
def buggy() {
    buggy()
}
buggy()
```

lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
svanoort: Yes -- worst case, calls can run deeper than they're supposed to, but since the limit we set is arbitrary that's not a huge problem.  We just want to cover the unbounded cases.

lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
svanoort: Varies depending on environment and settings.  In common cases, could be as little as 1000, or as much as 8000. 

In practice really all we care about is that "sane" code will run and that we catch runaway recursion -- also serializing the program will generate a StackOverflowError at some point (probably south of 1000 runs).

lib/src/test/groovy/com/cloudbees/groovy/cps/impl/FunctionCallBlockTest.groovy
svanoort: Fails implicitly, the above never returns "cheese" ;)

michaelneale:
nice sam!

<insert joke about solving the halting problem here>

svanoort:
@michaelneale Thanks!  The solution is rounding, of course if you squint really hard or use an imprecise numeric system,  1024 is close enough to Infinity.

michaelneale:
:bee:

svanoort:
@abayer  @jglick Please may I request re-review of this fix from one of you -- made some changes because with the exception simply thrown pipeline wasn't cleaning up the executor like it ought to have -- verification test in https://github.com/jenkinsci/workflow-cps-plugin/pull/183 -- thanks!

lib/src/test/groovy/com/cloudbees/groovy/cps/impl/FunctionCallBlockTest.groovy
michaelneale: i don’t think def thing is needed at all to show this anyway.

michaelneale:
Cool stuff Sam!

dgm-builder/pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>com.cloudbees</groupId>
         <artifactId>groovy-cps-parent</artifactId>
-        <version>1.21-SNAPSHOT</version>
+        <version>1.21-stop-recursion2-SNAPSHOT</version>
```
jglick: Revert please. Not necessary for testing downstream anyway; just use the timestamp in the dependency.

lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
```diff
@@ -5,8 +5,10 @@
  */
 public interface DepthTrackingEnv extends Env {
 
-    /** Limit on how deeply environments can recurse */
-    int MAX_LEGAL_DEPTH = 3000;
+    /** Limit on how deeply environments can recurse.
+     *  Capped somewhat low to try to limit the ability to run a program that will generate a StackOverflowError when serialized. */
+    int MAX_LEGAL_DEPTH = 1024;
```
jglick: I am afraid this might kill legitimate programs which just happen to use, say, a recursive function to process a list or whatever. (Yes common Groovy idioms would typically make that unnecessary, but…)

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsClosureDef.java
```diff
@@ -34,6 +35,19 @@ Next invoke(Env caller, SourceLocation loc, Object receiver, List<?> args, Conti
 
         // TODO: who handles 'it' ?
 
+        // 'e' will be a DepthTrackingEnv for all pipelines begun with this version of the library
+        // but if we upgrade groovy-cps and resume older pipelines, they won't be.
+        if (e instanceof DepthTrackingEnv && ((DepthTrackingEnv)e).getDepth() > DepthTrackingEnv.MAX_LEGAL_DEPTH) {
+            StackOverflowError err;
+            int depth = ((DepthTrackingEnv) e).getDepth();
+            if (loc != null) {
```
jglick: Maybe better to extract a utility function out of this (`@CheckForNull Next`).

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsClosureDef.java
```diff
@@ -34,6 +35,19 @@ Next invoke(Env caller, SourceLocation loc, Object receiver, List<?> args, Conti
 
         // TODO: who handles 'it' ?
 
+        // 'e' will be a DepthTrackingEnv for all pipelines begun with this version of the library
+        // but if we upgrade groovy-cps and resume older pipelines, they won't be.
```
jglick: Is that actually true? `DepthTrackingEnv` is just a marker interface. Existing instances of classes which retroactively implement it will be `instanceof DepthTrackingEnv` after a restart.

I think what you meant to say is that `getDepth()` would be 0 in such cases, since the field would be null after deserialization.

svanoort: I'm honestly not 100% sure if we'll have a path that provides an Env that isn't a DepthTrackingEnv -- but I'm also cautious about things like this.  We can revise the comment if you like.

dgm-builder/pom.xml
svanoort: It'll be reverted before merge -- I'm using this for local testing and don't seem to have the capability to deploy a SNAPSHOT -- and this isn't part of the repository-permissions-updater.

lib/src/main/java/com/cloudbees/groovy/cps/DepthTrackingEnv.java
michaelneale: yes, but necessary

svanoort: They'd run into a StackOverflowError eventually if they *weren't* NonCPS code though -- we're just setting an explicit threshold.  1024 is high enough that I feel comfortable saying "if you need more than this, you shouldn't be running it as part of CPS execution." 

There's always the escape valve of NonCPS annotations and ability to rewrite logic to use a non-recursive form.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsClosureDef.java
svanoort: You know what,  I don't know what I was thinking.  We don't need to worry about this at all, fixing it.

jglick:
> don't seem to have the capability to deploy a SNAPSHOT

Mentioned out of band but FTR:

```bash
alias deploysnapshot='mvn clean install source:jar deploy:deploy -DaltDeploymentRepository=maven.jenkins-ci.org::default::https://repo.jenkins-ci.org/snapshots/'
```

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsFunction.java
```diff
@@ -18,8 +19,18 @@ public CpsFunction(List<String> parameters, Block body) {
...
         assignArguments(args,e);
+
+        if (e.getDepth() > DepthTrackingEnv.MAX_LEGAL_DEPTH) {
+            StackOverflowError err;
+            if (loc != null) {
+                err = new StackOverflowError("Excessively nested closures/functions at "+loc+" - look for unbounded recursion - call depth: "+e.getDepth());
+            } else {
+                err = new StackOverflowError("Excessively nested closures/functions - look for unbounded recursion - call depth: "+e.getDepth());
+            }
+            return e.getExceptionHandler(StackOverflowError.class).receive(err);
```
jglick: Useful to extract a method for this stuff (DRY).

svanoort:
I'm holding off merge-release until I can figure out wtf is going on with the Windows issue in https://github.com/jenkinsci/workflow-cps-plugin/pull/183#issuecomment-337769172

svanoort:
Tests were just flaky (unrelated issues, addressed elsewhere) - so I'm going ahead to merge since this has been approved.

PR #79, opened 2017-10-20
[JENKINS-47363] Meaningful errors for oversized map/list expressions

abayer:
[JENKINS-47363](https://issues.jenkins-ci.org/browse/JENKINS-47363)

This is probably better than #78 - that works around the problem, but
in a fairly hackish way, while this instead accepts the limitation and
gives a meaningful error message when a list expression is >250
elements or a map expression is >125 entries.

cc @reviewbybees

### DIFF
--- a/lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
+++ b/lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
@@ -2,6 +2,7 @@ package com.cloudbees.groovy.cps
 
 import com.cloudbees.groovy.cps.impl.CpsCallableInvocation
 import groovy.transform.NotYetImplemented
+import org.codehaus.groovy.control.MultipleCompilationErrorsException
 import org.junit.Test
 import org.jvnet.hudson.test.Issue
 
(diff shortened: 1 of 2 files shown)


lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -877,4 +878,35 @@ def (c, d) = ['fourth']
...
+        } catch (Exception e) {
+            assert e instanceof MultipleCompilationErrorsException
+            assert e.message.contains('List expressions can only contain up to 250 elements')
+        }
+    }
+
+    @Issue("JENKINS-47363")
+    @Test
+    void excessiveMapElements() {
+        def l = 0..250
```
jglick: Not 125?

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -877,4 +878,35 @@ def (c, d) = ['fourth']
...
-}
+
+    @Issue("JENKINS-47363")
+    @Test
+    void excessiveListElements() {
+        def l = 0..250
+        def s = l.join(",\n")
+        try {
+            assert evalCPS("""
+def b = [${s}]
```
jglick: I would recommend first asserting that you _can_ successfully create a list of 250 elements (or even of every size up to and including 250); and only then verifying that an attempt to make one of 251 elements fails gracefully.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -877,4 +878,35 @@ def (c, d) = ['fourth']
...
+return b.size()
+""") == 251
+        } catch (Exception e) {
+            assert e instanceof MultipleCompilationErrorsException
```
jglick: I guess this was some other error type previously?

abayer: Correct.

abayer:
@svanoort Shall we go ahead with this sucker?

svanoort:
@abayer Merge away

PR #80, opened 2018-01-30
Do not run the sandbox transformer on field initializers created by the CPS transformer

jglick:
Finally tracked down a silly bug that was resulting in gratuitous `Checker.checkedStaticCall` invocations:

```groovy
private static final CpsFunction __cps__123 = __cps__123();
//                                            ^^^^^^^^^^^^
```

@reviewbybees


svanoort:
Previously reviewed out-of-band and merging now.

PR #81, opened 2018-02-20
[FIXED JENKINS-45575] Add support for multiple assignment in sandbox

abayer:
[JENKINS-45575](https://issues.jenkins-ci.org/browse/JENKINS-45575)

Turns out JENKINS-45575 is because I added proper support for multiple assignment in non-sandbox but not in sandbox. Oops.

Also get rid of `SandboxCpsTransformer#visitDeclarationExpression` by switching to overridden methods for getting the value for both single and multiple assignments that do the `SandboxTransformer#mightBePositionalArgumentConstructor` logic directly.

cc @reviewbybees


abayer:
@svanoort It should handle implicit type conversions the same way single assignment does - it definitely handles the `'foo' as File` case properly.

PR #82, opened 2018-02-22
[FIXED JENKINS-49679] Don't run RHS of multiple assignment more than once

abayer:
[JENKINS-49679](https://issues.jenkins-ci.org/browse/JENKINS-49679)

Evaluate the RHS into a temp variable and use that instead of the original expression.

cc @reviewbybees


lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -1103,6 +1107,7 @@ public void run() {
                 }
             });
         } else {
+            System.err.println("V: " + exp);
```
jglick: delete

PR #84, opened 2018-03-09
[FIXED JENKINS-50038] Fix "${foo}" for a CPS-transformed toString

abayer:
[JENKINS-50038](https://issues.jenkins-ci.org/browse/JENKINS-50038)

`GString#writeTo(Writer)` ends up calling (indirectly) `InvokerHelper#format(Object,boolean,int)`, which calls `toString()` on the value. Which means for CPS-transformed `toString()` methods, we've got...a CPS-transformed method being called in a non-CPS transformed context. Yay. So we need to cheat that and make sure the `toString()` call happens beforehand.

I don't *love* this. I'd prefer if we didn't have to always call `toString()` and would only do that if the object (or resulting object in the case of method calls) was CPS-transformed and its `toString()` was CPS-transformed...plus this is adding yet more bloat to the CPS-transformed classes/methods, leading to more opportunities for [JENKINS-37984](https://issues.jenkins-ci.org/browse/JENKINS-37984)...

cc @reviewbybees


abayer:
@jglick - like I said over on [this comment](https://issues.jenkins-ci.org/browse/JENKINS-31314?focusedCommentId=330622&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-330622), I'm not sure what the path forward there is. Pointers would be appreciated. =)

abayer:
Closing this in favor of the better logging effort.

PR #85, opened 2018-03-09
[JENKINS-31314] Diagnostics for mismatched function call vs. CpsCallableInvocation catch

jglick:
[JENKINS-31314](https://issues.jenkins-ci.org/browse/JENKINS-31314)


dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
```diff
@@ -389,6 +389,7 @@ public JExpression visitMethodInvocation(MethodInvocationTree mt, Void __) {
                                     .arg(n(e));
                             } else if (overloadsResolved.containsKey(overloadResolved)) {
                                 // Private, so delegate to our mangled version.
+                                // TODO add a String parameter to each internal helper method for the expected methodName to pass to CpsCallableInvocation.<init>
```
abayer: What do you mean by internal helper method here?

jglick: The mangled `$each__java_util_Iterator__groovy_lang_Closure` or whatever. It could be improved to take a parameter naming the method name under which we expect `methodCall` to be invoking it. Usually just `each`, but might be `$each__java_util_Iterable__groovy_lang_Closure` for the case that one DGM method is delegating to another.

jglick:
No longer a werk-in-progress but I have lost permission to remove that label.

lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java
```diff
@@ -53,18 +68,51 @@ protected Next methodCall(final Env e, final SourceLocation loc, final Continuat
...
 
+    private static Object findEffectiveReceiver(Object receiver, Map<Object, Boolean> encountered) {
+        if (!(receiver instanceof CpsClosure)) {
+            return receiver;
+        }
+        if (encountered == null) {
+            encountered = new IdentityHashMap<>();
+        }
+        if (encountered.put(receiver, true) == null) {
+            return findEffectiveReceiver(((CpsClosure) receiver).getOwner(), encountered);
```
dwnusbaum: Any idea of how many recursive calls we'd expect for most `CpsClosures`? I doubt we need to be worried about stack overflow for non-pathological cases but I'm curious if we should expect closer to 1, 10, or 100 recursive calls on average.

jglick: Should be one call per nested block in which a method call is made. This recursion was needed to handle cases like

```groovy
timestamps {
  node {
    ansicolor {
      something.broken()
    }
  }
}
```

PR #86, opened 2018-03-15
[JENKINS-44924] Add support for .sort {}, .unique {}, and friends.

abayer:
[JENKINS-44924](https://issues.jenkins-ci.org/browse/JENKINS-44924)

We handle these through a new `CpsDelegatingClosure` class, which takes a `CpsClosure`, calls it, and handles the resulting `CpsCallableInvocation` directly, without letting it bubble up to the calling context. This also prevents some false positives in #85.

(oh, and I also added a sanity test for `translatable.txt` inspired by the sanity tests for whitelists in `script-security`)

cc @reviewbybees


lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsDelegatingClosure.java
```diff
@@ -0,0 +1,94 @@
...
+    public Object call(Object... args) {
+        try {
+            cpsClosure.call(args);
+        } catch (CpsCallableInvocation inv) {
+            Continuable continuable = new Continuable(inv.asBlock());
+            Outcome o = continuable.run0(new Outcome(null, null), Continuable.categories);
+            if (o.isSuccess()) {
+                return o.getNormal();
+            } else {
+                throw new RuntimeException(o.getAbnormal());
```
abayer: I'm not sure if this is actually the right way to handle errors, so suggestions would be appreciated.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsDelegatingClosure.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+    }
+
+    @Override
+    public Object call(Object arguments) {
+        if (arguments.getClass().isArray()) {
```
abayer: As it turns out, this right here is why my previous attempts in this same general direction failed miserably: the method resolution for `call(1, 2, 3)` would end up hitting this instead of `call(Object...)`, resulting in us then calling `call(new Object[]{new Object[]{1,2,3}})`, effectively. That...messed things up.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsDelegatingClosure.java
```diff
@@ -0,0 +1,93 @@
...
+     * Here we approximate by using varargs.
+     * <p>
+     * {@link CurriedClosure} invokes this method directly (via {@link MetaClassImpl#invokeMethod(Class, Object, String, Object[], boolean, boolean)}
+     */
+    public Object doCall(Object... args) {
+        return call(args);
+    }
+
+    @Override
+    public boolean isCase(Object candidate){
```
abayer: Without the `isCase` override, this falls apart for comparator/case situations.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsDefaultGroovyMethodsTest.groovy
```diff
@@ -268,13 +264,11 @@ class CpsDefaultGroovyMethodsTest extends AbstractGroovyCpsTest {
...
             ["reverseEachMapEntry", "def r = ''; ['a': 1, 'b': 2, 'c': 3].reverseEach { e -> r += e.key }; return r", "cba"],
 
-            /* TODO would need to translate OrderBy and ClosureComparator
             // .sort
             ["sortList", "[3, 1, -2, -4].sort { i -> i * i }", [1, -2, 3, -4]],
             ["sortArray", "([3, 1, -2, -4] as Integer[]).sort { i -> i * i }", [1, -2, 3, -4]],
-            ["sortMapEntryByKey", "[a: 3, c: 1, b: -2, d: -4].sort { e -> e.key }", [a: 3, b: -2, c: 1, d: -4]],
-            ["sortMapEntryByValue", "[a: 3, c: 1, b: -2, d: -4].sort { e -> e.value }", [d: -4, b: -2, c: 1, a: 3]],
-            */
+            ["sortMapEntryByKey", "[a: 3, c: 1, b: -2, d: -4].sort { e -> e.key }.keySet().toList()", [a: 1, b:2, c: 3, d:4].keySet().toList()],
```
abayer: Turns out the map sort tests were passing even when they shouldn't have been - this change made them fail without the fix, so yay.

abayer:
Downstream PRs in `script-security` (for whitelisting all the new methods) and `workflow-cps` coming shortly.

abayer:
...joy. Turns out this doesn't quite work in all scenarios - running the `CpsDefaultGroovyMethodsTest` in `workflow-cps`, I'm seeing the `max(Map,Closure)` and `min(Map,Closure)` tests fail with `hudson.remoting.ProxyException: org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object 'org.jenkinsci.plugins.workflow.cps.Safepoint@6e4b1592' with class 'org.jenkinsci.plugins.workflow.cps.Safepoint' to class 'int'`.

abayer:
It also seems to have broken `takeWhile(String,Closure)`, which...confuses me at first glance.

abayer:
More specifically, `takeWhile(String,Closure)` works in `groovy-cps`'s tests, and fails in `workflow-cps`'s. Fun.

abayer:
Created https://github.com/jenkinsci/workflow-cps-plugin/pull/213 consuming this PR, fwiw.

dgm-builder/pom.xml
```diff
@@ -58,11 +58,30 @@
...
+                <artifactId>gmavenplus-plugin</artifactId>
+                <version>1.5</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>testGenerateStubs</goal>
+                            <goal>testCompile</goal>
+                        </goals>
+                        <configuration>
+                            <!-- bad defaults, fixed in Jenkins patch of gmaven-plugin -->
```
jglick: So why not just use that patch…?

dgm-builder/pom.xml
```diff
@@ -107,6 +126,18 @@
             <artifactId>guava</artifactId>
             <version>11.0.1</version>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.11</version>
```
jglick: 4.12 preferably

dgm-builder/src/main/resources/com/cloudbees/groovy/cps/tool/translatable.txt
```diff
@@ -1,133 +1,88 @@
-org.codehaus.groovy.runtime.DefaultGroovyMethods.<E,T,U,V>inject(E[],U,groovy.lang.Closure<V>)
```
jglick: Just resorting here? Would rather keep the original sort order to minimize diff and show that you are only adding entries.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsDelegatingClosure.java
jglick: `GroovyRuntimeException`? It is handled specially by something IIRC. Should be testable by having a closure throwing, say, `IOException` and then a `catch` block for that type.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsDelegatingClosure.java
```diff
@@ -0,0 +1,93 @@
...
+        try {
+            return cpsClosure.call(args);
+        } catch (CpsCallableInvocation inv) {
+            Continuable continuable = new Continuable(inv.asBlock());
+            Outcome o = continuable.run0(new Outcome(null, null), Continuable.categories);
+            if (o.isSuccess()) {
+                return o.getNormal();
+            } else {
+                throw new RuntimeException(o.getAbnormal());
+            }
```
dwnusbaum, line 63: I spent a bit of time looking into this and I think the reason the patch works here but not in `workflow-cps` is related to the safepoint mechanism. As it stands, this PR only executes the CPS-transformed closure up to the first safepoint. With a hack like the following, basic tests against `workflow-cps` start to pass:

```suggestion
            Outcome o = new Outcome(null, null);
            while (true) {
                o = continuable.run0(o, Continuable.categories);
                if (o.isSuccess()) {
                    Object normal = o.getNormal();
                    if (normal != null && normal.getClass().getName().equals("org.jenkinsci.plugins.workflow.cps.Safepoint")) {
                        o = new Outcome(null, null);
                    } else {
                        return normal;
                    }
                } else {
                    throw new RuntimeException(o.getAbnormal());
                }
            }
```

That said, I'm not sure how we could get something like `[-2, 1, -4, 3].sort({i -> sleep 1; i * i})` to execute in a way that makes sense using this approach. Maybe that's not a big deal though, since most closures used for these methods probably don't call Pipeline steps at all, let alone async Pipeline steps.

If I have some more time I will look into translating the methods that have issues into a variant that just throws an exception if they are called with a `CpsClosure` to at least make the current issues easier for users to understand.

jglick: > throws an exception if they are called with a `CpsClosure`

You do not think https://github.com/jenkinsci/workflow-cps-plugin/pull/211 would suffice?

dwnusbaum: That system works fine, but I think it would still be nice to have a more explicit fail-fast warning for cases where we know the method is not supported, e.g. `UnsupportedOperationException: DefaultGroovyMethods.sort(Foo, Closure) is not supported by Pipeline, see https://www.jenkins.io/doc/book/pipeline/cps-method-mismatches/ for details.`

dwnusbaum:
I think that the way that Groovy casts instances of `Closure` to preexisting types for these methods is essentially impossible to reconcile with the design of this library in a way that would work seamlessly in all cases. https://github.com/cloudbees/groovy-cps/pull/86#discussion_r586758345 describes some potential improvements we could make, as well as some limitations that could probably never be made to work and some issues with the approach that would need to be considered in more detail. In general though, no one is actively working on this, so I am going to close the PR.

PR #91, opened 2018-07-05
[JENKINS-52395] Add test case demonstrating stack overflow when a CPS transformed met…

stuartrowe:
…hod calls a super.method(...) which in turn calls super.methd(...)


stuartrowe:
@abayer Please review

abayer:
Thanks, @stuartrowe!

PR #92, opened 2018-08-21
Update to headers

aakoch:
It looks like maybe markdown changed since this was written. I added a space after # so the text was made into a header.

I didn't change the last line. I'm not sure why it shows a diff.


dwnusbaum:
> I didn't change the last line. I'm not sure why it shows a diff.

Perhaps the line endings changed?

PR #45, opened 2017-01-26
[JENKINS-31484] Translate field accesses to AttributeAccessBlock rather than PropertyAccessBlock

jglick:
[JENKINS-31484](https://issues.jenkins-ci.org/browse/JENKINS-31484)

@reviewbybees esp. @kohsuke

PR #95, opened 2019-02-22
[JENKINS-31484] Add ignored test for unfixed variant of JENKINS-31484

dwnusbaum:
See [JENKINS-31484](https://issues.jenkins-ci.org/browse/JENKINS-31484).

This PR adds an ignored test for another variant of that issue that works in regular Groovy but not in groovy-cps.


dwnusbaum:
> I thought we had fixed this long ago, but maybe it was some related syntax

A few lines up in `fieldViaGetter` there are tests for `x` and `this.@x` that you added in #45, so maybe that is what you are thinking of?

PR #96, opened 2019-03-29
com.cloudbees:cloudbees-oss-parent:9

jglick:
Analogous to https://github.com/jenkinsci/groovy-sandbox/pull/55 but picking up https://github.com/cloudbees/cloudbees-oss-parent/pull/6. @abayer @dwnusbaum @daniel-beck


jglick:
Note that I have no write permission here, so if you want it, merge it.

PR #101, opened 2019-08-13
JENKINS-58620 GroovyShell.evaluate should be skipped in the mismatch handler

sparshev:
Tests placed into jenkinsci/workflow-cps-plugin#315

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java
@@ -75,6 +75,8 @@ abstract class ContinuationGroup implements Serializable {
                     // CpsScript.invokeMethod e.g. on a UserDefinedGlobalVariable cannot be predicted from here.
                     expectedMethodNames.add("call");
                     laxCall = !((Script) effectiveReceiver).getBinding().getVariables().containsKey(methodName); // lax unless like invokePropertyOrMissing
+                } else if (effectiveReceiver instanceof GroovyShell && methodName.equals("evaluate")) {
+                    expectedMethodNames.add("run");
                 } else if (effectiveReceiver instanceof CpsBooleanClosureWrapper && methodName.equals("callForMap")) {
                     expectedMethodNames.add("call");
                 } else if ((effectiveReceiver instanceof ListWithDefault || effectiveReceiver instanceof MapWithDefault) && methodName.equals("get")) {


lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
```diff
@@ -63,6 +64,9 @@ void checkMismatch(Object expectedReceiver, List<String> expectedMethodNames) {
         if (expectedReceiver instanceof MetaClass && expectedMethodName.equals("invokeMethod")) {
             return; 
         }
+        if (expectedReceiver instanceof GroovyShell && expectedMethodName.equals("evaluate")) {
```
dwnusbaum: Does the following code in `ContinuationGroup.methodCall` not already handle this case? https://github.com/cloudbees/groovy-cps/blob/e930c15e010c658690a98631e57a0672f286beab/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java#L71-L74

Given your example mismatch warning seems to be `expected to call org.jenkinsci.plugins.workflow.cps.CpsGroovyShell.evaluate but wound up catching X.run`, it seems like it might be better to adapt that condition slightly than to add the exclusion here. Maybe you just need to add (`|| expectedReceiver instanceof GroovyShell`) to the condition?

sparshev: Tried to add as you suggested - but if fails later on [ContinuationGroup.java:77](https://github.com/cloudbees/groovy-cps/blob/e930c15e010c658690a98631e57a0672f286beab/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java#L77) with
```
java.lang.ClassCastException: org.jenkinsci.plugins.workflow.cps.CpsGroovyShell cannot be cast to groovy.lang.Script
```

But looks like working if I'm adding it as an additional if - and looks better than just skipping the thing.

dwnusbaum: Thanks for looking into it!

sparshev:
@jglick could you please check it? What do you think - it's good enough to squash & merge?

sparshev:
Ok, one commit now. @abayer, @jglick @dwnusbaum please merge if it looks ok.

sparshev:
@jglick any updates?

abayer:
Looks good, merging.

sparshev:
Great, thank you)

PR #102, opened 2019-08-19
[FIXED JENKINS-58407] Treat closure fields and map values properly

abayer:
[JENKINS-58407](https://issues.jenkins-ci.org/browse/JENKINS-58407)

Don't throw a mismatch for cases like `foo.someField()` where `foo` is
either an `Object` with a field `someField` that's a closure, or a
`Map` with a `Closure` value for the key `someField`.

https://github.com/jenkinsci/workflow-cps-plugin/pull/316

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
@@ -15,9 +15,7 @@ import javax.annotation.CheckForNull;
 import groovy.lang.Closure;
 import groovy.lang.GroovyObject;
 import groovy.lang.MetaClass;
-import groovy.lang.MetaProperty;
 import groovy.lang.MissingPropertyException;
-import org.codehaus.groovy.runtime.metaclass.MissingPropertyExceptionNoStack;
 
 /**
  * When an CPS-interpreted method is invoked, it immediately throws this error
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
@@ -16,6 +16,8 @@ import groovy.lang.Closure;
 import groovy.lang.GroovyObject;
 import groovy.lang.MetaClass;
 import groovy.lang.MetaProperty;
+import groovy.lang.MissingPropertyException;
+import org.codehaus.groovy.runtime.metaclass.MissingPropertyExceptionNoStack;
 
 /**
  * When an CPS-interpreted method is invoked, it immediately throws this error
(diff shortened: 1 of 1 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
```diff
@@ -64,6 +69,19 @@ void checkMismatch(Object expectedReceiver, List<String> expectedMethodNames) {
             return; 
         }
 
+        if (expectedReceiver instanceof GroovyObject) {
+            Object propVal = ((GroovyObject) expectedReceiver).getMetaClass().getProperty(expectedReceiver, expectedMethodName);
```
dwnusbaum: I thought about something like this, but what if the implementations of `MetaClass.getProperty` or `Map.get` have side effects? I guess we just assume that in most cases they won't, so this is safe?

abayer: Not that we'd care here, I don't think. `MetaClass.getProperty` is definitely clean, `Map.get`...well, if it's not clean, then there are a bunch of other things that can go wrong outside of this scope. =)

abayer: This isn't perfect, but it does deal with these scenarios for mismatch purposes.

dwnusbaum: Fair enough. It just reminds me a lot of SECURITY-1465 (https://github.com/jenkinsci/groovy-sandbox/commit/e30cd28d7b30cd606e78c22174cb04e0450244a7), where we ran into issues calling user-defined methods in places that expected "normal" behavior. I don't think it would be a security issue here since the actual method would still be sandbox-transformed, it would just cause confusing behavior.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
```diff
@@ -9,8 +9,15 @@
...
+import java.util.Map;
 import javax.annotation.CheckForNull;
-import groovy.lang.MetaClass; 
+
+import groovy.lang.Closure;
+import groovy.lang.GroovyObject;
+import groovy.lang.MetaClass;
+import groovy.lang.MetaProperty;
+import groovy.lang.MissingPropertyException;
+import org.codehaus.groovy.runtime.metaclass.MissingPropertyExceptionNoStack;
```
dwnusbaum: Unused?

abayer: whoops, fixed.

lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
```diff
@@ -64,6 +71,23 @@ void checkMismatch(Object expectedReceiver, List<String> expectedMethodNames) {
...
+                    return;
+                }
+            } catch (MissingPropertyException mpe) {
+                // Do nothing - this just means the property didn't exist.
+            }
+        }
+        if (expectedReceiver instanceof Map && ((Map) expectedReceiver).containsKey(expectedMethodName)) {
+            Object mapVal = ((Map) expectedReceiver).get(expectedMethodName);
+            if (mapVal instanceof Closure) {
+                return;
```
dwnusbaum: nit: I guess to be extra specific we could check that `"call".equals(methodName)` here and in the other condition before returning, or do we intentionally want to ignore closures entirely? I'm not really sure if `methodName` could be something other than `call` in this case, or if we gain anything by being more specific.

abayer: If you're calling `foo.someField()`, where `someField` is a closure, you're calling `foo.someField.call()` - but it wouldn't hurt. I'll add something.

PR #103, opened 2019-08-29
Suppress mismatch warnings for invocations of Continuable.suspend

dwnusbaum:
`Continuable.suspend` explicitly changes control flow in a way that we can't track, so we shouldn't warn about it.

CC @alexanderrtaylor


dwnusbaum:
I realized that this PR doesn't make sense. Any code running into problems because it uses `Continuable.suspend` should switch to the non-deprecated overload of that method and pass the method name that will be used on the Groovy side to invoke the code that calls `Continuable.suspend` to avoid any warnings.

PR #108, opened 2020-05-06
[JENKINS-62064] Make assignment expressions evaluate to their RHS

dwnusbaum:
See [JENKINS-62064](https://issues.jenkins-ci.org/browse/JENKINS-62064).

Fixes a case where groovy-cps semantics did not match Groovy semantics.

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/AssignmentBlock.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/AssignmentBlock.java
@@ -64,10 +64,10 @@ public class AssignmentBlock extends CallSiteBlockSupport {
         }
 
         /**
-         * Just straight assignment from RHS to LHS, then done
+         * Assign from RHS to LHS and pass RHS to the continuation.
          */
         public Next assignAndDone(Object rhs) {
-            return lhs.set(rhs,k);  // just straight assignment
+            return lhs.set(rhs, then(new ConstantBlock(rhs), e, k));
         }
 
         /**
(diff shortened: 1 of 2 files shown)


lib/src/main/java/com/cloudbees/groovy/cps/impl/AssignmentBlock.java
```diff
@@ -64,10 +64,10 @@ public Next fixLhs(Object lhs) {
         }
 
         /**
-         * Just straight assignment from RHS to LHS, then done
+         * Assign from RHS to LHS and pass RHS to the continuation.
          */
         public Next assignAndDone(Object rhs) {
-            return lhs.set(rhs,k);  // just straight assignment
+            return lhs.set(rhs, then(new ConstantBlock(rhs), e, k));
```
dwnusbaum, line 70: Instead of passing the current continuation `k` to `LValue.set`, which would result in `k.receive(null)` being called after `lhs` is set to `rhs`, we construct a new continuation that calls `k.receive(rhs)` internally and pass it to `lhs.set` instead.

I think we could also add a new method like `setAndGet` to the `LValue` interface and implement it in the 2-3 subclasses of `LValue`, but this seemed simpler and I didn't see any other callers of `LValue.set` that would want the new behavior.

lib/src/test/java/com/cloudbees/groovy/cps/CpsTransformerTest.java
```diff
@@ -965,30 +965,30 @@ public void spreadMapExpression() throws Throwable {
 
     @Test
     public void initialExpressionsInMethodsAreCpsTransformed() throws Throwable {
-        assertEquals(Boolean.FALSE, evalCPS(
+        assertEvaluate(Boolean.FALSE,
```
dwnusbaum, line 968: I should have made these changes as part of https://github.com/cloudbees/groovy-cps/commit/9a3b1b042211e83b48fea6a15096e2575b397a0b, but I did not notice it until I was resolving the merge conflict with the newly added test at the bottom of the file.

daniel-beck (approved):
Looks reasonable. Untested.

PR #109, opened 2020-05-07
[JENKINS-57253] Verify correct usage of break and continue statements during compilation

dwnusbaum:
See [JENKINS-57253](https://issues.jenkins-ci.org/browse/JENKINS-57253) and https://github.com/jenkinsci/workflow-cps-plugin/pull/356 (CC @basil).

Makes improper use of `break` or `continue` statements fail during compilation with the same error messages as Groovy itself instead of at runtime with a CPS-specific error message.

`LabelVerifier` normally runs during `CompilePhase.CLASS_GENERATION`, but CPS transformation runs earlier during `CompilePhase.CANONICALIZATION` and obliterates the original control-flow statements that `LabelVerifier` would check, so the `LabelVerifier` that runs during `CompilePhase.CLASS_GENERATION` can only catch issues in groovy-cps itself. 

To be able to catch errors in the user's program, this PR makes `CPSTransformer` run `LabelVerifier` manually prior to performing the CPS transformation. Looking at `LabelVerifier`, I do not think there are problems with running it during an earlier compilation phase.

### DIFF
--- a/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
@@ -30,6 +30,7 @@ import org.codehaus.groovy.classgen.GeneratorContext;
 import org.codehaus.groovy.classgen.Verifier;
 import org.codehaus.groovy.control.CompilePhase;
 import org.codehaus.groovy.control.Janitor;
+import org.codehaus.groovy.control.LabelVerifier;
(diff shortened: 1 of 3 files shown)


basil:
The amount of value produced by this community that remains held up in unmerged and/or unreleased PRs is, from an outside perspective, staggering.

olamy (approved):

car-roll (approved):

PR #117, opened 2021-12-01
Demonstrating GROOVY-9608 method interception

basil:
Demonstrating that apache/groovy@cf51a3f fixes [GROOVY-9608](https://issues.apache.org/jira/browse/GROOVY-9608) without altering method interception.


basil:
Note that `mvn clean verify -Dtest=com.cloudbees.groovy.cps.sandbox.SandboxInvokerTest` passes in both commits cf. jenkinsci/jenkins#5939.

PR #119, opened 2022-08-31
Trying to build on JDK 11

jglick:
First problem was that `javax.annotation.*` is gone.

Bigger problem is

```
Exception in thread "main" java.lang.NoClassDefFoundError: com/sun/tools/javac/file/ZipArchive
	at com.cloudbees.groovy.cps.tool.Driver.run(Driver.java:48)
	at com.cloudbees.groovy.cps.tool.Driver.main(Driver.java:28)
Caused by: java.lang.ClassNotFoundException: com.sun.tools.javac.file.ZipArchive
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 2 more
```

which is not solved by

```bash
--add-exports jdk.compiler/com.sun.tools.javac=ALL-UNNAMED
```

(Maybe not right; just Googling for solutions.)

But using Shade does not work either:

```
Exception in thread "main" java.lang.NullPointerException
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.Locations.getPathEntries(Locations.java:148)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.Locations.getPathEntries(Locations.java:133)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.Locations.access$000(Locations.java:70)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.Locations$BootClassPathLocationHandler.computePath(Locations.java:616)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.Locations$BootClassPathLocationHandler.lazy(Locations.java:642)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.Locations$BootClassPathLocationHandler.getLocation(Locations.java:576)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.Locations.getLocation(Locations.java:677)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.JavacFileManager.getLocation(JavacFileManager.java:803)
	at com.cloudbees.groovy.cps.com.sun.tools.javac.file.JavacFileManager.list(JavacFileManager.java:617)
[...]

PR #120, opened 2022-10-03
Update dgm-builder to be able to build and run against Java 9+ and switch to Jenkins plugin parent POM

dwnusbaum:
Replaces #119. Various compiler-related APIs changed between Java 8 and Java 9, which combined with the Jigsaw module changes mean that you can only compile and run `dgm-builder` using Java 8.

This PR updates `dgm-builder` to be able to be built and run against at least Java 17 and Java 11. It probably works with Java 9 and all newer versions, but I have not checked. **With this PR, `dgm-builder` can no longer be built or run using Java 8**, but the `groovy-cps` library can still be run using Java 8.

This PR also switches the parent POM to a recent version of https://github.com/jenkinsci/plugin-pom. The reason to do this now is that after merging this PR, I am going to merge this library into https://github.com/jenkinsci/workflow-cps-plugin, and switching the parent in advance will make the merge simpler.

TODO:
* [x] Test against workflow-cps
    * Everything seems to work fine, but I saw one test failure that I have been unable to reproduce:  `CpsFlowDefinition2Test.endlessRecursionNonCPS` failed because it timed out, and at the time of the timeout there was a 15k-line stack trace evaluating the Groovy script (which normally throws a `StackOverflowError` with around 1000 lines for me). I don't see any obvious link between that test and the changes here.
* [x] Confirm that `groovy-cps` can still be _run_ using Java 8
* [x] Confirm what versions of Java are able to build/run `dgm-builder` when using this PR
    * At least Java 11.0.17 and 17.0.4.1 work, which should be good enough
* [x] Investigate usage of unexported types in `jdk.compiler` module


dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Driver.java
```diff
@@ -1,10 +1,6 @@
 package com.cloudbees.groovy.cps.tool;
 
 import com.sun.codemodel.writer.FileCodeWriter;
-import com.sun.tools.javac.api.JavacTool;
-import com.sun.tools.javac.file.JavacFileManager;
-import com.sun.tools.javac.file.RelativePath.RelativeDirectory;
-import com.sun.tools.javac.file.ZipArchive;
```
dwnusbaum, line 7: This class no longer exists as far as I can tell.

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
```diff
@@ -95,7 +95,7 @@
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
-import javax.annotation.Generated;
+import javax.annotation.processing.Generated;
```
dwnusbaum, line 98: IDK what the exact history is here, but this class is now an exported class in `java.compiler` and the source code is the same. The `RetentionPolicy` is `SOURCE`, so I think that changing the class is fine, but I'm not sure.

lib/pom.xml
```diff
@@ -45,6 +45,10 @@
                         <sourceRoot>${project.build.directory}/generated-sources/dgm</sourceRoot>
                         <executable>java</executable>
                         <arguments>
+                            <!-- Translator currently downcasts to non-exported types in a few cases, mainly to check types of identifiers, and there is no obvious alternative. -->
```
dwnusbaum, line 48: I think that some cases could be rewritten but the new code would be much more verbose. I am not sure though, and it would take some time to investigate further.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
```diff
@@ -186,7 +186,7 @@ class CpsTransformerTest extends AbstractGroovyCpsTest {
             } catch(StringIndexOutOfBoundsException e) {
               return e.message;
             }
-        """)=="String index out of range: -2"
+        """)=="begin 5, end 3, length 3"
```
dwnusbaum, line 189: The message changed in Java 9.

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Driver.java
```diff
@@ -45,10 +42,13 @@ public void run(File dir) throws Exception {
...
-            ZipArchive a = new ZipArchive((JavacFileManager) fileManager, new ZipFile(groovySrcJar));
-
-            for (String name : fileNames) {
-                src.add(a.getFileObject(new RelativeDirectory("org/codehaus/groovy/runtime"),name+".java"));
+            for (JavaFileObject jfo : fileManager.list(StandardLocation.CLASS_PATH, "org.codehaus.groovy.runtime", Collections.singleton(JavaFileObject.Kind.SOURCE), true)) {
+                for (String name : fileNames) {
+                    if (jfo.toUri().toString().endsWith("/org/codehaus/groovy/runtime/" + name + ".java")) {
+                        src.add(jfo);
+                        break;
+                    }
```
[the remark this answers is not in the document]
jglick: It has been years since I used this API so I do not recall either.

lib/pom.xml
jglick: This is just a build-time tool, it should not matter.

pom.xml
```diff
@@ -16,7 +16,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <groovy.version>2.4.7</groovy.version>
+        <groovy.version>2.4.21</groovy.version>
```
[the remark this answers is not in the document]
jglick: Matches https://github.com/jenkinsci/jenkins/pull/5939 anyway.

jglick (approved):
Nice!

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
```diff
@@ -357,26 +351,28 @@ public JExpression visitMethodInvocation(MethodInvocationTree mt, Void __) {
                                 .arg(n(mst.getIdentifier()));
                     }
                 } else
-                if (ms instanceof JCIdent) {
+                if (ms instanceof IdentifierTree) {
```
jglick, line 354: BTW acc. to docs you are never supposed to use `instanceof` with these APIs; rather you check some type enum and then cast. I doubt it matters in practice.

lib/pom.xml
dwnusbaum: True, but I looked into it in [3f0928b](https://github.com/cloudbees/groovy-cps/pull/120/commits/3f0928b84c126f3ceb1b5fdc728fc6de6a826d96) and migrating to only use the public API is relatively straightforward. In general the most common issue was using `JCIdent` to get the fully qualified class name of an identifier and/or its type, but we can switch to using `Element.toString()` and `Element.asType()`. (Awkward to use `toString`, but we were using `Symbol.toString()` previously, and there does not seem to be a better way, so...)

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
dwnusbaum: Ok, we could check `Tree.getKind()` instead and then cast but IDK what the difference would be. There were maybe 2-3 places with similar code before my changes. Do you have a link to the docs though? I see such a warning for `Element` but not for `Tree`: https://docs.oracle.com/en/java/javase/16/docs/api/java.compiler/javax/lang/model/element/Element.html.

jglick: Maybe this only applies to `Element` and not `Tree` then.

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Driver.java
basil: As of [JDK-8059976](https://bugs.openjdk.org/browse/JDK-8059976), as I discovered in https://github.com/jenkins-infra/backend-extension-indexer/pull/41.

lib/src/test/groovy/com/cloudbees/groovy/cps/CpsTransformerTest.groovy
dwnusbaum: Actually, for this specific overload of `substring`, [this commit](https://github.com/openjdk/jdk17/commit/c0b8844dce2335531d47eeaae8f6de3a48c5c2b6) which appears to have first landed in JDK 14 is what changes the message.

That makes things a bit awkward, since we want to be able to build using at least Java 11 and Java 17, but we can adapt the test to support either version.

lib/src/test/java/com/cloudbees/groovy/cps/CpsTransformerTest.java
```diff
@@ -172,16 +174,19 @@ public void workflowCallingWorkflow() throws Throwable {
...
-            "}\n");
+        try {
+            evalCPS(
+                "def foo() {\n" +
+                "  'abc'.substring(5);\n" + // will caught exception
+                "  return 'fail';\n" +
+                "}\n" +
+                "return foo()\n");
+            fail("Should have failed");
+        } catch(StringIndexOutOfBoundsException e) {
```
dwnusbaum, line 185: Oh I guess this test specifically wanted to check the behavior of try/catch, so I will refactor this.

dwnusbaum: https://github.com/cloudbees/groovy-cps/pull/120/commits/c1a2ff4c70527eb315a75c1946a7f0fe460d9c06

dgm-builder/src/main/java/com/cloudbees/groovy/cps/tool/Translator.java
dwnusbaum: Turns out that it is actually simpler to depend on `javax.annotation-api` and use the old package for now while we are running `dgm-builder` against Java 11 but compiling and running `groovy-cps` (and `workflow-cps` once we merge this into it) for compatibility with Java 8 to avoid needing to specify a bunch of pom overrides.

dgm-builder/pom.xml
```diff
@@ -102,17 +79,6 @@
...
             <version>${groovy.version}</version>
         </dependency>
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>11.0.1</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci</groupId>
-            <artifactId>test-annotations</artifactId>
```
dwnusbaum, line 112: The parent POM adds this dependency automatically.

lib/src/main/java/com/cloudbees/groovy/cps/Next.java
```diff
@@ -141,7 +141,7 @@ public static Next terminate0(Outcome v) {
     /**
      * As a {@link Continuation}, just ignore the argument.
      */
-    public Next receive(Object _) {
+    public Next receive(Object unused) {
```
dwnusbaum, line 144: These changes are also not strictly necessary right now, but will be needed to be able to compile against Java 11 in the future, and this seems like as good a time as any to make the change.

lib/pom.xml
```diff
@@ -53,41 +47,12 @@
...
-            <version>1.0</version>
-          </dependency>
-        </dependencies>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.20</version>
-        <configuration>
-          <trimStackTrace>false</trimStackTrace> <!-- SUREFIRE-1226 workaround -->
```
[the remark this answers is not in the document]
jglick: Right.

lib/src/main/java/com/cloudbees/groovy/cps/Next.java
jglick: Is there a reason you are not just going to 11 now?

dwnusbaum: If we generate Java 11 bytecode for `groovy-cps`, then `workflow-cps` would require Java 11, and my understanding is that not many (if any?) plugins require Java 11 right now, and I am not really interested in experimenting with a widely-installed plugin like `workflow-cps`.

pom.xml
```diff
@@ -16,17 +18,14 @@
...
-        <incrementals-enforce-minimum.version>1.1</incrementals-enforce-minimum.version>
-        <incrementals-plugin.version>1.1</incrementals-plugin.version>
-        <incrementals.url>https://repo.jenkins-ci.org/incrementals/</incrementals.url>
         <scmTag>HEAD</scmTag>
     </properties>
 
-    <scm>
+    <scm child.scm.connection.inherit.append.path="false" child.scm.developerConnection.inherit.append.path="false" child.scm.url.inherit.append.path="false">
         <connection>scm:git:git@github.com/cloudbees/groovy-cps.git</connection>
         <developerConnection>scm:git:ssh://git@github.com/cloudbees/groovy-cps.git</developerConnection>
```
basil, line 30: SSH URLs are [deprecated](https://github.blog/2021-09-01-improving-git-protocol-security-github/). The following has been the most reliable replacement I have found:

```suggestion
        <developerConnection>scm:git:git@github.com:cloudbees/groovy-cps.git</developerConnection>
```

pom.xml
```diff
@@ -16,17 +18,14 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
basil, line 20: This is [already set in the plugin parent POM](https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml#L59).

```suggestion
```

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
```diff
@@ -115,7 +114,7 @@ public CpsTransformer() {
         super(CompilePhase.CANONICALIZATION);
     }
 
-    public void setConfiguration(@Nonnull TransformerConfiguration config) {
+    public void setConfiguration(@NonNull TransformerConfiguration config) {
```
basil, line 117: Seems to me that if we are migrating some classes in `lib/src/main` from JSR 305 to SpotBugs annotations, we should migrate all of them. It seems the following changes from basil#1 did not make it into the final version of this PR:

```java
diff --git a/lib/src/main/java/com/cloudbees/groovy/cps/Env.java b/lib/src/main/java/com/cloudbees/groovy/cps/Env.java
index 0555aa9..f38c3d6 100644
--- a/lib/src/main/java/com/cloudbees/groovy/cps/Env.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/Env.java
@@ -3,9 +3,9 @@ package com.cloudbees.groovy.cps;
 import com.cloudbees.groovy.cps.impl.LocalVariableBlock;
 import com.cloudbees.groovy.cps.impl.VariableDeclBlock;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import groovy.lang.Closure;
-import javax.annotation.CheckForNull;
 import java.io.Serializable;
 import java.util.List;
 
diff --git a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
index ff2423e..ab8b3fd 100644
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CallEnv.java
@@ -7,7 +7,7 @@ import com.cloudbees.groovy.cps.Next;
 import com.google.common.collect.Maps;
 
-import javax.annotation.Nullable;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
diff --git a/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java b/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java
index d44e234..fd0c4ed 100644
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/ContinuationGroup.java
@@ -7,6 +7,7 @@ import com.cloudbees.groovy.cps.Env;
 import com.cloudbees.groovy.cps.Next;
 import static com.cloudbees.groovy.cps.impl.SourceLocation.*;
+import edu.umd.cs.findbugs.annotations.CheckReturnValue;
 import groovy.lang.GroovyCodeSource;
 import groovy.lang.GroovyShell;
 import groovy.lang.ListWithDefault;
@@ -21,7 +22,6 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
-import javax.annotation.CheckReturnValue;
 import org.codehaus.groovy.runtime.InvokerInvocationException;
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
 import org.codehaus.groovy.runtime.callsite.CallSite;
diff --git a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
index 9b2ed80..68fa8ef 100644
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/CpsCallableInvocation.java
@@ -5,12 +5,13 @@ import com.cloudbees.groovy.cps.Continuation;
 import com.cloudbees.groovy.cps.Env;
 import com.cloudbees.groovy.cps.Next;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+
 import java.util.List;
 
 import static java.util.Arrays.*;
 import java.util.Collections;
 import java.util.Map;
-import javax.annotation.CheckForNull;
 
 import groovy.lang.Closure;
 import groovy.lang.GroovyObject;
diff --git a/lib/src/main/java/com/cloudbees/groovy/cps/impl/TryBlockEnv.java b/lib/src/main/java/com/cloudbees/groovy/cps/impl/TryBlockEnv.java
index 60cadc8..1f484f7 100644
--- a/lib/src/main/java/com/cloudbees/groovy/cps/impl/TryBlockEnv.java
+++ b/lib/src/main/java/com/cloudbees/groovy/cps/impl/TryBlockEnv.java
@@ -5,7 +5,7 @@ import com.cloudbees.groovy.cps.Continuation;
 import com.cloudbees.groovy.cps.Env;
 import com.cloudbees.groovy.cps.Next;
 
-import javax.annotation.Nullable;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
```

lib/src/main/java/com/cloudbees/groovy/cps/Next.java
basil: Full support for Java 11 in the plugin toolchain will need to wait a few more months until 2.361 LTS and future are more widely adopted.

basil (approved):
Very nice!

lib/src/main/java/com/cloudbees/groovy/cps/CpsTransformer.java
dwnusbaum: Oops, good catch, added in [e623ff9](https://github.com/cloudbees/groovy-cps/pull/120/commits/e623ff9584c3f46f36192c30ec5234d8e6e6215d).

pom.xml
dwnusbaum: Removed in [e623ff9](https://github.com/cloudbees/groovy-cps/pull/120/commits/e623ff9584c3f46f36192c30ec5234d8e6e6215d).

pom.xml
dwnusbaum: Thanks, modified in [e623ff9](https://github.com/cloudbees/groovy-cps/pull/120/commits/e623ff9584c3f46f36192c30ec5234d8e6e6215d).

jglick (approved):
Great!

lib/src/main/java/com/cloudbees/groovy/cps/Next.java
jglick: > `workflow-cps` would require Java 11

Well, yes, a 2.361.x baseline.

> not many (if any?) plugins require Java 11 right now

Actually there are a fair number already on 2.361.x (or even subsequent weeklies), mainly because of core UI changes.

PR #121, opened 2022-10-19
Delete unused APIs

dwnusbaum:
The `com.cloudbees.groovy.cps.green` package as well as a few `Continuable` methods and related classes do not appear to be used outside of this library in the Jenkins ecosystem, and it seems pointless to keep them around. Additionally, `Continuable.fork` is broken and does not work correctly.

Each commit is independent. 9849734ca2e71fe9a911f431f1f14698c1427017 and 6420ea566dd5c2e8f046693d4cbe2186ee2769bf seem relatively uncontroversial - the package could be recreated outside of the library if someone is using this library outside of Jenkins (I hope not!) and needs the package, and `Continuable.fork` just doesn't work correctly.

15e728eae40ee9442f089a95af18d8781c0d0b05 and ca79d4068cf1df6f6fbe79b77e1d82576431115f delete methods that as far as I know are not broken, but seem to be unused in the Jenkins ecosystem. They could not easily be recreated outside of the library because they deal with `Continuable` internals, so if we want to be extra careful I could revert those commits.


lib/src/main/java/com/cloudbees/groovy/cps/Continuable.java
```diff
@@ -185,26 +166,6 @@ public Outcome call() {
         });
     }
 
-    /**
-     * Sets a super-interrupt.
-     *
-     * <p>
-     * A super interrupt works like {@link Thread#interrupt()} that throws
-     * {@link InterruptedException}. It lets other threads interrupt the execution
-     * of {@link Continuable} by making it throw an exception.
```
jglick, line 194: Huh. I guess this was not in fact used to implement the **Pause** function in Pipeline?

lib/src/main/java/com/cloudbees/groovy/cps/Continuable.java
```diff
@@ -244,64 +205,6 @@ public List<StackTraceElement> getStackTrace() {
         return r;
     }
 
-    /**
-     * Ignore whatever that we've been doing, and jumps the execution to the given continuation.
-     *
-     * <p>
-     * Aside from the obvious use case of completely overwriting the state of {@link Continuable},
-     * more interesting case is
```
jglick, line 252: …? Hmm.

jglick (approved):
Deleting dead code, great.

lib/src/main/java/com/cloudbees/groovy/cps/Continuable.java
dwnusbaum: Yeah, this was a bit surprising to me. I think that the safepoint mechanism is currently all that is being used. We could keep this around it case it could be used from `CpsThread` to optimize the time between initiating a pause and the program actually pausing, but we could always bring it back if there is some reason to work on that in the future.

jglick: YAGNI

PR #122, opened 2022-10-25
Migrate all Groovy test sources to Java

dwnusbaum:
The Groovy tests in this library have always caused issues for me. They cannot be run individually from my IDE when I need to debug a specific test, and I often run into subtle issues with incremental compilation and stub generation that are difficult to debug. I have not found Groovy's `assert` statement to be particularly helpful when debugging tests, and so in my opinion the main benefit of the Groovy sources was the readability improvement from using multi-line strings for Groovy scripts, which I am more than willing to give up to make the codebase more consistent and to be able to remove the dependency on `gmavenplus`.

The diff is essentially unreviewable, but I verified that the expected number of tests are running before and after the change:

Before:

```
[WARNING] Tests run: 422, Failures: 0, Errors: 0, Skipped: 4
```

After:

```
[WARNING] Tests run: 422, Failures: 0, Errors: 0, Skipped: 13
```

The +9 skipped tests are from tests that previously used Groovy's `@NotYetImplemented` annotation and now use `@Ignore`.

I did notice a few unusual things while converting everything. For example, [CpsStringGroovyMethods.evalCpsAndSync](https://github.com/cloudbees/groovy-cps/blob/0011f8fb050c1dd6ca48406ec4075176d45e6ea8/lib/src/test/groovy/com/cloudbees/groovy/cps/CpsStringGroovyMethodsTest.groovy#L92-L99) has code like `evalCPS(script) == value` with no `assert`, oops! `evalCPS` has internal assertions that check that the actual CPS-transformed result and actual non-CPS-transformed result match, but some of the tests expected a different value, and needed to be modified (e.g. the actual return value of the scripts in [CpsStringGroovyMethodsTest.find](https://github.com/cloudbees/groovy-cps/blob/0011f8fb050c1dd6ca48406ec4075176d45e6ea8/lib/src/test/groovy/com/cloudbees/groovy/cps/CpsStringGroovyMethodsTest.groovy#L25-L31) is `boo`, not `raboof`).


jglick (approved):
Great!

